PR #8, opened 2016-05-03
Updated graph API error handling for timing issue

jamarsto:
Updated graph API error handling for timing issue on new user registration with the Amazon and Google identity providers.

isUserInRole will now return false if invoked in the landing page and there is an IO Error when accessing the graph API.  In reality isUserInRole should not really be used in the successful login landing page, but returning false is better than showing a 500 error.


lenala:
looks good, merging.

PR #10, opened 2016-05-10
updates to clean up unused variables and publish path for eclipse integration and clear authorisation cache on logout

jamarsto:


jamarsto:
Additional updates to flush authorisation cache on logout

PR #376, opened 2017-05-08
Implement client telemetry for the Intellij plugin

juniwang:
The changes include:
- Add AzureDialogWrapper as the base class for all dialogs where telemetryies are sent by default. Any dialog who wants to enable telemetry should inherit from this class. Existing dialogs are replaced with new base class already.
- Add AzureWizardStep for WizardStep. Similar as above.
- And AzureAnAction to be the base class for all AnActions.
- AzureAbstractHandler for all AbstractHandler implementations.
- Refactor and move the AppInsights client to the core project so that it can be reuesd by both plugins and the telelmetry name/properties are formatted.

Other minor changes:
- Updat gitignore to ignore those auto-downloaded dependencies.
- Refactor azure blob/table/queue editor to enable the telemetry in a consistent way


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/serverexplore/hdinsightnode/BlobContainerNode.java
```diff
@@ -23,17 +23,19 @@
...
 import com.microsoft.tooling.msservices.model.storage.ClientStorageAccount;
 import com.microsoft.tooling.msservices.serviceexplorer.Node;
 import com.microsoft.tooling.msservices.serviceexplorer.NodeActionEvent;
 import com.microsoft.tooling.msservices.serviceexplorer.NodeActionListener;
 
+import java.util.HashMap;
 import java.util.Map;
 
-public class BlobContainerNode extends Node {
+public class BlobContainerNode extends Node implements TelemetryProperties{
```
adashen: We touched hdinsight code?

Utils/azuretools-core/src/com/microsoft/azuretools/telemetry/TelemetryProperties.java
```diff
@@ -0,0 +1,12 @@
+package com.microsoft.azuretools.telemetry;
```
adashen: License header needed for this file?

adashen:
Please delete sqljdbc42.jar. The jar cannot submit into public repo

juniwang:
the code will not compile without `sqljdbc42.jar`, at least on my local computer it fails.  Will the CI be broken if remove it from the repo?

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/serverexplore/hdinsightnode/BlobContainerNode.java
juniwang: Not meant to

vscjavaci:

vscjavaci:
ok to test

Utils/azuretools-core/src/com/microsoft/azuretools/telemetry/TelemetryProperties.java
juniwang: Updated

PR #378, opened 2017-05-09
Revert "remove module common and hdinsight from intellij plugin"

voidfunction:
Reverts Microsoft/azure-tools-for-java#377


vscjavaci:

PR #380, opened 2017-05-09
remove ant build file and unnecessary class file from hdinsight

voidfunction:


vscjavaci:

PR #404, opened 2017-05-11
quick fix the NullPointerException

juniwang:
The host and  hostVM field is null before the VM is really created.


juniwang:
Cannot get the subId in this case which is a field of the hostVM that is null before it's really created. And it will be created after the Next/Finish is clicked. We definitely can get the subId by other means, but there is only a quick fix for the blocking issue.

PR #413, opened 2017-05-11
a real fix for the telemetry while adding a new docker VM host

juniwang:
add the subscriptionID and name to the telemetry properties for both scenarios: create new VM host and select existing VM host


voidfunction:

PR #425, opened 2017-05-15
Add parent pom in Utils

adashen:
1. Add a parent pom under "Utils" 
2. The parent pom declare the dependencymanagement to avoid jar conflict
3. Besides, it is also an aggregation pom to build the three utils
4. Please consider add dependencymanagement when introduce new dependency in theses util modules
5. Besides, fix a class loading error of telementry. Expose the package in manifest


voidfunction:

Issue #428, opened 2017-05-16
[Intellij.2017.Spark.Template] Run plugin with IC version (community version), hit error: org.jetbrains.idea.maven.project doesn't exist.


wezhang:
```
D:\src\github\Microsoft\my\azure-tools-for-java\PluginsAndFeatures\azure-toolkit-for-intellij\src\com\microsoft\azure\hdinsight\projects\samples\MavenProjectGenerator.java
Error:(36, 40) java: 程序包org.jetbrains.idea.maven.project不存在
Error:(143, 9) java: 找不到符号
  符号:   类 MavenProjectsManager
  位置: 类 com.microsoft.azure.hdinsight.projects.samples.MavenProjectGenerator
Error:(143, 40) java: 找不到符号
  符号:   变量 MavenProjectsManager
  位置: 类 com.microsoft.azure.hdinsight.projects.samples.MavenProjectGenerator
```

ziyel:
@wezhang 
Hi Wei, please add below lines to file `PluginsAndFeatures/azure-toolkit-for-intellij/azure-toolkit-for-intellij.iml` to fix this temporarily. I will figure out how to fix this permanently.
```xml
<orderEntry type="module-library" scope="PROVIDED">
   <library>
     <CLASSES>
       <root url="jar://$APPLICATION_HOME_DIR$/plugins/maven/lib/maven.jar!/" />
     </CLASSES>
     <JAVADOC />
     <SOURCES />
   </library>
</orderEntry>
```

ziyel:
Fix in https://github.com/Microsoft/azure-tools-for-java/pull/429.

ziyel:
Fix in #429

PR #429, opened 2017-05-16
add maven plugin lib to intellij iml file to fix the local debug

ziyel:
Fix issue #428.


vscjavaci:

PR #430, opened 2017-05-16
rename the intellij ui widgets according to the naming convention

jdneo:


voidfunction:

jdneo:
@adashen @testforstephen @Eskibear

PR #431, opened 2017-05-16
add IDE version to telemetry property

juniwang:


voidfunction:

adashen:
Is the telemetry can get the plugin version, like "3.0.2"?

juniwang:
plugin version is already included in previous implementation

Issue #435, opened 2017-05-16
[Eclipse+IntelliJ] Properties function issue


bsaby:
Redis Cache workflow:
[Eclipse +IntelliJ2017 +IntelliJ2016]
The Properties function has not beenimplemented, so the Properties menu should not appear
![image](https://cloud.githubusercontent.com/assets/22470312/26090646/e952c4da-3a38-11e7-9630-51173121c22a.png)
@jdneo

adashen:
@bsaby fix merged please use the latest build for verification

bsaby:
Test passed.
Build used:
artifacts/azure-toolkit-for-intellij.2016.develop.daily.4.05-17-2017.zip
artifacts/azure-toolkit-for-intellij.2017.develop.daily.4.05-17-2017.zip
artifacts/WindowsAzurePlugin4EJ.develop.daily.4.05-17-2017.zip

PR #444, opened 2017-05-16
remove properties action in intellij

jdneo:
#435 
Remove the list properties action for Redis Cache in Intellij


jdneo:
@adashen @Eskibear @testforstephen @bsaby

PR #447, opened 2017-05-16
Split SparkSubmissionContentPanel from SparkSubmissionExDialog

djlan:
Split SparkSubmissionContentPanel from SparkSubmissionExDialog so the Spark submission UI can be reused

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/spark/run/configuration/RemoteDebugSettingsEditor.java
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/spark/run/configuration/RemoteDebugSettingsEditor.java
@@ -24,11 +24,27 @@ package com.microsoft.azure.hdinsight.spark.run.configuration;
 
 import com.intellij.openapi.options.ConfigurationException;
 import com.intellij.openapi.options.SettingsEditor;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.project.ex.ProjectManagerEx;
+import com.microsoft.azure.hdinsight.common.CallBack;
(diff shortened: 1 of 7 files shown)


vscjavaci:

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/spark/run/configuration/RemoteDebugSettingsEditor.java
```diff
@@ -42,6 +58,17 @@ protected void applyEditorTo(@NotNull RemoteDebugRunConfiguration remoteDebugRun
     @NotNull
     @Override
     protected JComponent createEditor() {
-        return new JPanel();
+
+        Project project = myRunConfiguration.getProject();
+        List<IClusterDetail> cachedClusterDetails = ClusterManagerEx.getInstance().getClusterDetailsWithoutAsync(true, project);
```
voidfunction: >getClusterDetailsWithoutAsync [](start = 83, length = 29)

getClusterDetailsWithoutAsync is just getting all cached clusters. Is it on purpose?

voidfunction:
:shipit:

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/spark/run/configuration/RemoteDebugSettingsEditor.java
djlan: It is not on purpose. However, we do want this dialog to pop up promptly. I will open an issue so we can get the no cluster bug fixed. Then this won't be an issue.

---
In reply to: [116899228](https://github.com/Microsoft/azure-tools-for-java/pull/447#discussion_r116899228) [](ancestors = 116899228)

Issue #440, opened 2017-05-16
[IntelliJ] Redis Cache: Resorse group list can not be really refreshed


bsaby:
[IntelliJ 2017 + IntelliJ2016]
After Redis Cache node refresh, the Use existing resourse group list can not be really refreshed.
The group that was recently deleted will not disappear in the list, and the newly created group will not appear in the list.
![image](https://cloud.githubusercontent.com/assets/22470312/26091727/96aa6ae2-3a3f-11e7-8700-081e4cda20f2.png)

@jdneo

adashen:
fixed please verify

bsaby:
Test passed.
Builds used:
artifacts/azure-toolkit-for-intellij.2016.develop.daily.4.05-17-2017.zip
artifacts/azure-toolkit-for-intellij.2017.develop.daily.4.05-17-2017.zip

PR #452, opened 2017-05-17
Fix resgrp refresh

jdneo:
fix bug #440


voidfunction:

jdneo:
@adashen @Eskibear @testforstephen @bsaby

PR #456, opened 2017-05-18
fix telemetry bug

jdneo:


voidfunction:

PR #462, opened 2017-05-18
fix bug of spark job submission on Eclipse

voidfunction:


voidfunction:

Issue #458, opened 2017-05-18
Redis Cache Workflow: UI issue


bsaby:
[Eclipse]
Build: 
artifacts/WindowsAzurePlugin4EJ.develop.52.05-18-2017.zip

**Windows**
![image](https://cloud.githubusercontent.com/assets/22470312/26185247/4f6a77c0-3bbc-11e7-8a4b-fff8f0659877.png)
**Mac**
![image](https://cloud.githubusercontent.com/assets/22470312/26185265/6dee9b68-3bbc-11e7-9755-f590ebcf1f72.png)
**Linux**
![image](https://cloud.githubusercontent.com/assets/22470312/26185302/b9eedd70-3bbc-11e7-8d0f-f6e562dc7433.png)
EXPECT(for subscription )
![image](https://cloud.githubusercontent.com/assets/22470312/26185325/d442932e-3bbc-11e7-81ae-8469392391fb.png)

jdneo:
@bsaby fixed, please verify in the new build, thanks.

bsaby:
Repro in Linux.
Build used: artifacts/WindowsAzurePlugin4EJ.develop.56.05-18-2017.zip.
![image](https://cloud.githubusercontent.com/assets/22470312/26197227/0fea010a-3bf4-11e7-8062-51bbeee9674c.png)

adashen:
@bsaby , please verify use the latest build

bsaby:
Still repro in Linux.
Build used: artifacts/WindowsAzurePlugin4EJ.develop.65.05-19-2017.zip

![image](https://cloud.githubusercontent.com/assets/22470312/26239551/1c05fe78-3cb0-11e7-80a1-0dbf5ba70ea4.png)

bsaby:
Build: Eclipse Daily 27
![image](https://user-images.githubusercontent.com/22470312/26861072-fbeeaed2-4b73-11e7-8b4a-95ca4863cfa6.png)

PR #464, opened 2017-05-18
fix some form issue

jdneo:
1. change the form in eclipse
2. make the form in intellij can be resizable
#458


voidfunction:

PR #468, opened 2017-05-18
auto generate manifest.mf & copy dependencies for com.microsoft.azure…

Eskibear:
…tools.sdk

Previously we copy dependencies of `azuretools-core`, `azure-explorer-common`, `hdinsight-node-common` **RESPECTIVELY** into `PluginsAndFeatures/AddLibrary/AzureLibraries/com.microsoft.azuretools.sdk/dependencies`, which introduces jar conflicts and extra jar files(e.g. old version, test scope dependencies).

Now we copy the dependences of above 3 packages together, and accordingly auto-generate file `META_INF/MANIFEST.MF`.

Note that, the build squence changes a bit:
```
# preparing 
mvn install -f Utils/pom.xml
mvn install -f PluginsAndFeatures/AddLibrary/AzureLibraries/pom.xml

# build eclipse-plugin or intellij-plugin
...
```


voidfunction:

voidfunction:
Thanks. Seems we should add module com.microsoft.azuretools.sdk.lib to https://github.com/Microsoft/azure-tools-for-java/blob/develop/Utils/pom.xml#L39

Eskibear:
@voidfunction yes adding module has the same results as `mvn install -f PluginsAndFeatures/AddLibrary/AzureLibraries/pom.xml`. BUT it's not that natural to have ` com.microsoft.azuretools.sdk.lib` being a module of `Utils`. What do you think?

PluginsAndFeatures/AddLibrary/AzureLibraries/pom.xml
```diff
@@ -0,0 +1,145 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>com.microsoft.azuretools</groupId>
+    <artifactId>com.microsoft.azuretools.sdk.lib</artifactId>
+    <version>3.0.3</version>
+    <packaging>bundle</packaging>
+    <name>Package for Microsoft Azure Libraries for Java Plugin</name>
+    <organization><name>Microsoft Corp.</name></organization>
+    <parent>
```
voidfunction: parent statement here seems inappropriate if module om.microsoft.azuretools.sdk.lib is not part of Utils.

Eskibear: <parent> statement is used to inherit <dependencyManagement> so as to avoid jar conflict.  Do you have other better solutions or suggestions?

voidfunction: let's just keep it if we don't have a better way to handle it. It looks more clearly now. Thanks.

Eskibear:
For better user experience, close this PR, and use  [a workaround PR](https://github.com/Microsoft/azure-tools-for-java/pull/473) instead.

PR #481, opened 2017-05-19
fix null pointer bug

ziyel:


ziyel:
Fix the https://github.com/Microsoft/azure-tools-for-java/issues/486

PR #516, opened 2017-05-22
Fix BuildDevint script to modify Intellij plugin version

testforstephen:


voidfunction:

PR #525, opened 2017-05-22
fix telemetry 1000549

adashen:
use handler simple class name instead of text as event name

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/utils/AzureAbstractHandler.java
+++ b/PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/utils/AzureAbstractHandler.java
@@ -39,7 +39,7 @@ public abstract class AzureAbstractHandler extends AbstractHandler {
 
 		final Map<String, String> properties = new HashMap<>();
 		try {
-			Command cmd = event.getCommand();
+			final Command cmd = event.getCommand();
 			String handlerName = this.getClass().getSimpleName();
 			
 			if (cmd != null) {
--- a/PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/utils/AzureAbstractHandler.java
+++ b/PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/utils/AzureAbstractHandler.java
@@ -23,10 +23,10 @@ import java.util.HashMap;
 import java.util.Map;
 
 import org.eclipse.core.commands.AbstractHandler;
+import org.eclipse.core.commands.Command;
 import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.core.commands.ExecutionException;
 import org.eclipse.core.commands.common.NotDefinedException;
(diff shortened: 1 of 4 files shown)


PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/utils/AzureAbstractHandler.java
```diff
@@ -39,12 +39,19 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
...
+			Command cmd = event.getCommand();
+			String handlerName = this.getClass().getSimpleName();
 			
-			AppInsightsClient.createByType(AppInsightsClient.EventType.Action, event.getCommand().getName(),	null, properties);
+			if (cmd != null) {
+				properties.put("CategoryId", cmd.getCategory().getId());
+				properties.put("Category", cmd.getCategory().getName());
+				properties.put("CommandId", cmd.getId());
+				properties.put("Text", cmd.getName());
+				if(null == handlerName || handlerName.isEmpty()) {
```
juniwang: `this` means the current object so it will NEVER be null. Same for its class name. I do think this check is unnecessary.

PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/utils/AzureAbstractHandler.java
```diff
@@ -39,12 +39,19 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
 
 		final Map<String, String> properties = new HashMap<>();
 		try {
-			properties.put("CategoryId", event.getCommand().getCategory().getId());
-			properties.put("Category", event.getCommand().getCategory().getName());
-			properties.put("CommandId", event.getCommand().getId());
-			properties.put("Text", event.getCommand().getName());
+			Command cmd = event.getCommand();
```
juniwang: Suggest adding `final` keyword as possible as you can to prevent unintended changes. It's one of the best practices in java.

PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/utils/AzureAbstractHandler.java
adashen: java anonymous class will have empty value of simple classname. So just add a check to take care of such edge case

PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/utils/AzureAbstractHandler.java
adashen: sure

PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/utils/AzureAbstractHandler.java
juniwang: Got it , you are right.

Issue #485, opened 2017-05-19
[Spark-Debug] The Job configuration - Dynamic Configuraiton's component of Spark Debug Run Configuration editor only show '...'


wezhang:
![image](https://cloud.githubusercontent.com/assets/7385667/26241387/0788fb56-3cb7-11e7-84e2-3362b585eaeb.png)

djlan:
Fix checked in. Please verify.

wezhang:
Fixed, thanks!

PR #530, opened 2017-05-22
Fix issue 485 where Dynamic Configuration becomes invisible in Submit…

djlan:
This is to fix issue #485


vscjavaci:

PR #533, opened 2017-05-22
Add functions of loading Spark job config from property file. 

djlan:
Merge dynamic config and file config into one where file config loaded will update config table.


vscjavaci:

PR #536, opened 2017-05-23
Fix 'RUN' executor is un-usable for Spark debug configuration

wezhang:


vscjavaci:

PR #561, opened 2017-05-24
fix bug: getting driver log error on job view

voidfunction:


voidfunction:
test case failed due to mocked livy service time out. @wezhang Could you help take a look? Thx

PR #567, opened 2017-05-25
fix bug: getting driver log error on job view

voidfunction:


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/task/YarnHistoryTask.java
```diff
@@ -40,29 +44,32 @@
...
+        HtmlPage htmlPage = WEB_CLIENT.getPage(path);
 
-        HttpEntity httpEntity = response.getEntity();
+        // parse pre tag from html response
+        // there's only one 'pre' in response
+        DomNodeList<DomElement> preTagElements = htmlPage.getElementsByTagName("pre");
 
-        return IOUtils.toString(httpEntity.getContent());
+        if (preTagElements.size() == 0) {
+            return "No logs here or logs not available";
```
wezhang: Should throws an exception here?

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/task/YarnHistoryTask.java
```diff
@@ -40,29 +44,32 @@
 
     protected final IClusterDetail clusterDetail;
     protected final String path;
-    private final CredentialsProvider credentialsProvider =  new BasicCredentialsProvider();
-
+    private static final CredentialsProvider CREDENTIALS_PROVIDER =  new BasicCredentialsProvider();
```
wezhang: Why static here?

Utils/hdinsight-node-common/resources/htmlResources/hdinsight/job/html/sparkjobview.js
```diff
@@ -265,7 +265,7 @@ function getSparkDriverLog() {
         executorsObject = JSON.parse(s);
         var hostPort = getDriverPortFromExecutor(executorsObject);
         ipAddress = hostPort.split(":")[0];
-        var url = localhost + projectId + "/jobhistory/logs/" + ipAddress + "/port/30050/" + containerId + "/" + containerId + "/spark/stderr?restType=yarnhistory";
+        var url = localhost + projectId + "/jobhistory/logs/" + ipAddress + "/port/30050/" + containerId + "/" + containerId + "/livy/stderr?restType=yarnhistory";
```
wezhang: I remember the driverUrl can get from Livy response rather than combin it that is always changed without notifications.

PluginsAndFeatures/azure-toolkit-for-intellij/build.gradle
```diff
@@ -78,6 +78,7 @@ dependencies {
     testCompile 'org.assertj:assertj-swing-junit:3.5.0'
     runtime 'com.microsoft.azure:azure-client-runtime:1.0.2'
     runtime 'com.microsoft.azure:azure-client-authentication:1.0.2'
+
```
wezhang: Whitespace, the file can be removed from the change

Utils/hdinsight-node-common/resources/htmlResources/hdinsight/job/html/sparkjobview.js
voidfunction: Job view didn't have a livy response here. Only active sessions and batches have such response.

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/task/YarnHistoryTask.java
voidfunction: only one instance is required here since we will set the username/password in constructor

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/task/YarnHistoryTask.java
voidfunction: we need to return something in html web page. I think empty content can be expected here.

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/task/YarnHistoryTask.java
wezhang: Instance would be better here, it would help the testing in future. And the static limits that all yarn history task instance would share the same basic type credential provider.

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/task/YarnHistoryTask.java
wezhang: The method already throws exceptions, it might be better to construct an error html web page at the outside catching function.

PR #568, opened 2017-05-25
Fix the issue of SparkConfigures type conversion

wezhang:


vscjavaci:

Issue #575, opened 2017-05-27
[Intellij Spark debug] Clusters can not display on debug configuration dialog for the first time 


ruoxinCheng:
![image](https://cloud.githubusercontent.com/assets/20041888/26517270/be964522-42c7-11e7-943a-3deb598a3222.png)

ruoxinCheng:
Verified it.

PR #577, opened 2017-05-27
Fix the cluster refreshing is blocked by the model dialog

wezhang:
Introduced the replay subject for toolwindow messages.


vscjavaci:

wezhang:
To fix #575

PR #581, opened 2017-05-27
perf improve: telemetry client singleton

### LINKED ISSUE #524: App insight object is created everytime when sending telemetry
The problem is at both eclipse and intellij. Just take eclipse as example. In the file AppInsightsCustomEvent.java, an object of class [TelemetryClient] is created every time when calling [create] for sending telemetry. Creating this object and [setInstrumentationKey()] for it is very time consuming: on my machine the latency is longer than 5 seconds. As a result, every time we send a metric to AI, there is such a performance penalty. A solution could be to create a global object rather than a temporary one in each call, if so, then be aware to add proper "lock" logic to avoid resource collision.

The counterpart file in IntelliJ is still called AppInsightsCustomEvent.java.

Eskibear:
fix #524

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/telemetry/AppInsightsCustomEvent.java
+++ b/PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/telemetry/AppInsightsCustomEvent.java
@@ -90,8 +90,7 @@ public class AppInsightsCustomEvent {
     }
 
     public static void createFTPEvent(String eventName, String uri, String appName, String subId) {
-        TelemetryClient telemetry = new TelemetryClient();
-        telemetry.getContext().setInstrumentationKey(key);
+        TelemetryClient telemetry = TelemetryClientSingleton.getTelemetry(key);
 
         Map<String, String> properties = new HashMap<String, String>();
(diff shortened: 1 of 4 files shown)


PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/telemetry/AppInsightsCustomEvent.java
```diff
@@ -27,94 +27,97 @@
...
+        }
+        synchronized(TelemetryClientSingleton.class){
+            telemetry.trackEvent(Messages.telemetryDenyAction, properties, null);
+            telemetry.flush();
+        }
+    }
+
+    public static void createFTPEvent(String eventName, String uri, String appName, String subId) {
+        TelemetryClient telemetry = new TelemetryClient();
+        telemetry.getContext().setInstrumentationKey(key);
```
jdneo: also `TelemetryClient telemetry = TelemetryClientSingleton.getTelemetry(key);` here?

Eskibear: yes, fixed now.

Eskibear:
retest this please

PR #605, opened 2017-06-08
rename some method names to make them more correct

jdneo:


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/helpers/UIHelperImpl.java
```diff
@@ -270,7 +270,7 @@ public String promptForOpenSSLPath() {
     }
 
     @Override
-    public void openView(String sid, String id) {
+    public void openRedisPropertyView(String sid, String id) {
```
Eskibear: general or dedicated?

jdneo: It's dedicated now, will try to generalize it when we need to show more kind of property views

PR #614, opened 2017-06-09
Fix intelliJ regression bug

adashen:
Fix the application insight regression bug. 
The reason of the bug is
1. Jar version is hard coded in the file list.
2. When jar is upgraded, it will fail to find the target jar.
Solution
1. Only store target jar name without version.
2. When find the target jar in the file folder, only compare with file name without version


PluginsAndFeatures/azure-toolkit-for-intellij/Test/java/com/microsoft/intellij/ui/libraries/AddLibraryUtilityTest.java
```diff
@@ -0,0 +1,24 @@
+package com.microsoft.intellij.ui.libraries;
```
jdneo: lisence

adashen: done

PR #617, opened 2017-06-09
Redis property intellij

jdneo:
Intellij can show redis cache's property


PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.azureexplorer/src/com/microsoft/azuretools/azureexplorer/helpers/UIHelperImpl.java
```diff
@@ -254,7 +254,7 @@ public boolean isDarkTheme() {
     }
     
     @Override
-    public void openRedisPropertyView(String sid, String id) {
+    public void openRedisPropertyView(Object project, String sid, String id, String type, String iconName) {
```
adashen: how about just use Node?

Issue #651, opened 2017-06-14
[Intellij] Clusters are double in azure explorer


ruoxinCheng:
![image](https://user-images.githubusercontent.com/20041888/27118356-788c6bde-510d-11e7-9efd-48c085b1f0a2.png)

ziyel:
@ruoxinCheng , please help to test.

ruoxinCheng:
Verified daily build, it fixed.

Issue #590, opened 2017-06-05
[HDInsight] Non-ASCII characters in comments


Eskibear:
Non-blocking issue. Low priority.
```
[INFO] --- maven-compiler-plugin:3.3:compile (default-compile) @ hdinsight-node-common ---
[INFO] Changes detected - recompiling the module!
[WARNING] File encoding has not been set, using platform encoding GBK, i.e. build is platform dependent!
[INFO] Compiling 112 source files to C:\Users\Sechs\Documents\Eskibear\azure-tools-for-java\Utils\hdinsight-node-common\target\classes
[ERROR] /C:/Users/Sechs/Documents/Eskibear/azure-tools-for-java/Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/serverexplore/hdinsightnode/StorageAccountNode.java:[67,22] unmappable character for encoding GBK
[ERROR] /C:/Users/Sechs/Documents/Eskibear/azure-tools-for-java/Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/serverexplore/hdinsightnode/StorageAccountNode.java:[67,23] unmappable character for encoding GBK
[INFO] /C:/Users/Sechs/Documents/Eskibear/azure-tools-for-java/Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/spark/common/SparkSubmissionParameter.java: Some input files use unchecked or unsafe operations.
[INFO] /C:/Users/Sechs/Documents/Eskibear/azure-tools-for-java/Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/spark/common/SparkSubmissionParameter.java: Recompile with -Xlint:unchecked for details.
```

Non-ASCII characters in comments, refer to [This line](https://github.com/Microsoft/azure-tools-for-java/blob/develop/Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/serverexplore/hdinsightnode/StorageAccountNode.java#L67)

ziyel:
Fix in https://github.com/Microsoft/azure-tools-for-java/pull/656

PR #656, opened 2017-06-14
remove duplicate spark clusters from rest api

ziyel:
+ Fix https://github.com/Microsoft/azure-tools-for-java/issues/590
+ Fix https://github.com/Microsoft/azure-tools-for-java/issues/651
The root cause is the clusters list return from api `https://main.hdinsight.ext.azure.com/api/Clusters/GetAll` is doubled. We temporarily fix the issue by  removing the duplicated clusters in the list.

### DIFF
--- a/Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/cluster/ClusterManager.java
+++ b/Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/cluster/ClusterManager.java
@@ -81,7 +81,7 @@ public class ClusterManager {
      * @return detailed cluster info list with specific cluster type
      * @throws AggregatedException
      */
-    public synchronized List<IClusterDetail> getHDInsightCausersWithSpecificType(
+    public synchronized List<IClusterDetail> getHDInsightClustersWithSpecificType(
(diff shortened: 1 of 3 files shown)


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/cluster/ClusterManager.java
```diff
@@ -79,26 +81,31 @@ private ClusterManager() {
...
             } else {
                 if (clusterDetail.getType().equals(type)) {
-                    filterClusterDetailList.add(clusterDetail);
+                    filterClusterDetailMap.put(clusterDetail.getName(), clusterDetail);
                 }
             }
         }
 
+        List<IClusterDetail> filterClusterDetailList = new ArrayList<>();
+        for (IClusterDetail clusterDetail : filterClusterDetailMap.values()) {
```
voidfunction: filterClusterDetailList.addAll(filterClusterDetailMap.values())

ziyel: Thanks Long. Will fix in next PR.

PR #671, opened 2017-06-16
Split applicationinsight test key from prod

adashen:
The default application insight key in the code is for test. The Split strategy is as follows
1. Set the test key in pom/gradle and use resource filtering. So the default value will got the pom/gradle properties
2. When release to product. We will override the property through mvn/gradle command parameters. 
3. We will modify the jenkins job to do the release build and set the product key


Utils/azuretools-core/src/com/microsoft/azuretools/telemetry/TelemetryClientSingleton.java
```diff
@@ -3,25 +3,19 @@
 
 public final class TelemetryClientSingleton {
     private TelemetryClient telemetry = null;
-    private static String appInsightsKey = null;
 
     private static final class SingletonHolder {
         private static final TelemetryClientSingleton INSTANCE = init();
         private static TelemetryClientSingleton init(){
-            return new TelemetryClientSingleton(appInsightsKey);
+            return new TelemetryClientSingleton();
```
Eskibear: simply 
` private static final TelemetryClientSingleton INSTANCE = new TelemetryClientSingleton();`

adashen: will modify in next pull request

Issue #523, opened 2017-05-22
[Eclipse 3.0.1] Cannot show clusters after signing sucessful


ruoxinCheng:
![image](https://cloud.githubusercontent.com/assets/20041888/26297525/c6fdfee6-3f05-11e7-8d2e-7e44a8e13f76.png)

voidfunction:
Fixed.

ziyel:
Fix in #686
@ruoxinCheng , could you please help to verify in the latest daily build.

ruoxinCheng:
Yes, verified on Eclipse mars and Neon, it works

PR #686, opened 2017-06-23
set the read timeout for RestServiceManagerBaseImpl class

ziyel:


ziyel:
Fix https://github.com/Microsoft/azure-tools-for-java/issues/523
#### Root Cause
If we don't set the `readTimeout` for `HttpsURLConnection`, the `readTimeout` will finally set in blow codes of `sun.net.NetworkClient`
```Java
public void setReadTimeout(int var1) {
    if(var1 == -1) {
        var1 = defaultSoTimeout;
    }
    ...
}
```
And the `defaultSoTimeout` is a `protected static` variable.
```Java
protected static int defaultSoTimeout;
```
It maybe set by other codes out of our control, thus cause calling HDInsight restful api timeout.
As our test, the `defaultSoTimeout` was set to `1000`(1 second) when the issue occurred on Eclipse of Windows. In Intellij and the other no issue scenarios, the `defaultSoTimeout` is set to `0`(infinity).
So we just set the `readTimeout` to `0` for quick fixing the issue.
In future, we need to find a suitable timeout depends on the SLA of HDInsight restful api.

PR #688, opened 2017-06-26
Property refactor

jdneo:
- Use rxJava in Redis Property Presenter
- Add a shut down listerner in Eclipse property view, to make sure the view will be closed when eclipse will be shut down
- Change the resource release logic into lifecycle method: dispose(), instead of add a close listener

### DIFF
--- a/Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/rediscache/RedisPropertyViewPresenter.java
+++ b/Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/rediscache/RedisPropertyViewPresenter.java
@@ -61,6 +61,9 @@ public class RedisPropertyViewPresenter<V extends RedisPropertyMvpView> extends 
         })
         .subscribeOn(Schedulers.io())
         .subscribe(redis -> {
+            if (redis == null) {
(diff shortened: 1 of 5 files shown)


Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/rediscache/RedisPropertyViewPresenter.java
```diff
@@ -57,20 +56,19 @@ public void onGetRedisProperty(String sid, String id) {
...
-        }
-        RedisCacheProperty property = new RedisCacheProperty(redis.name(), redis.type(), redis.resourceGroupName(),
-                redis.regionName(), sid, redis.redisVersion(), redis.sslPort(), redis.nonSslPort(),
-                redis.keys().primaryKey(), redis.keys().secondaryKey(), redis.hostName());
-        ((RedisPropertyMvpView) getMvpView()).showProperty(property);
+        Observable.fromCallable(() -> {
+            return AzureMvpModelHelper.getInstance().getRedisCache(sid, id);
+        })
+        .subscribeOn(Schedulers.io())
+        .subscribe(redis -> {
```
adashen: check null case

PR #736, opened 2017-07-06
change the default error handler method in mvpview

jdneo:


Utils/azuretools-core/src/com/microsoft/azuretools/core/mvp/ui/base/MvpView.java
```diff
@@ -22,15 +22,13 @@
 
 package com.microsoft.azuretools.core.mvp.ui.base;
 
-import javax.swing.JOptionPane;
-
 public interface MvpView {
     
     default void onError(String message) {
-        JOptionPane.showMessageDialog(null, null, message, JOptionPane.ERROR_MESSAGE, null);
+        MvpUIHelperFactory.getInstance().getMvpUIHelper().showError(message);;
```
adashen: getMvpUIHelper could return null

PR #756, opened 2017-07-11
retire sun.misc.IOUtils

Eskibear:
related feature: azure-explorer


Eskibear:
It says `error: no suitable method found for readFully(InputStream,int)`, but the method exists since version 2.5 (exactly the version we use) according to [JavaDoc](https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/IOUtils.html#readFully(java.io.InputStream,%20int))
Original error log:
```
00:13:59.090 [ERROR] [system.err] /Users/sechs/Documents/Eskibear/azure-tools-for-java/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/helpers/storage/BlobExplorerFileEditor.java:638: error: no suitable method found for readFully(InputStream,int)
00:13:59.091 [ERROR] [system.err]                                             String errResponse = new String(IOUtils.readFully(errorStream, -1));
00:13:59.091 [ERROR] [system.err]                                                                                    ^
00:13:59.092 [ERROR] [system.err]     method IOUtils.readFully(Reader,char[]) is not applicable
00:13:59.092 [ERROR] [system.err]       (argument mismatch; InputStream cannot be converted to Reader)
00:13:59.092 [ERROR] [system.err]     method IOUtils.readFully(InputStream,byte[]) is not applicable
00:13:59.092 [ERROR] [system.err]       (argument mismatch; int cannot be converted to byte[])
```

Eskibear:
The build will pass if we explicitly add dependency of `commons-io:2.5` in `gradle.build`. Just wondering why should we do that, the commons-io should have been introduced in building Utils.

Issue #779, opened 2017-07-25
[IntelliJ] Stop process, the debug application still running


ruoxinCheng:

ruoxinCheng:
Verified it.

PR #795, opened 2017-07-27
Fix Spark Remote Debug STOP button not working issue

wezhang:


wezhang:
to fix #779

Issue #814, opened 2017-08-03
Need check sign in status before show deploy webapp dialog


adashen:

adashen:
fixed

PR #817, opened 2017-08-03
Web app deploy

jdneo:
1. change the method signature for onRefresh()
2. fix a issue: #814 
3. leverage the MavenProject.getPackaging() to get the file type name
4. implement some logic of init the run configuration dialog
5. remove the process listener (the listener has been added in the processhandler's constructor)


jdneo:
retest this please

PR #835, opened 2017-08-07
Web app deploy

jdneo:
can recover the settings after the user open the setting panel


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/ui/webapp/deploysetting/WebAppSettingPanel.java
```diff
@@ -444,56 +487,79 @@ public void actionPerformed(AnActionEvent anActionEvent) {
...
+            if (Comparing.equal(plan.id(), defaultServicePlan)) {
+                cbExistAppServicePlan.setSelectedItem(plan);
             }
         }
     }
 
     @Override
-    public void fillSubscription(List<Subscription> subscriptions) {
-        subscriptions.forEach(cbSubscription::addItem);
+    public void fillLocation(List<Location> locations) {
```
adashen: cannot make sure about the sequence that fillXX will be called before resetfrom. So pass configuration into this class

PR #839, opened 2017-08-07
eclipse publish to WAOL: popup err message if no docker daemon

Eskibear:


PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.container/src/com/microsoft/azuretools/container/Constant.java
```diff
@@ -55,4 +55,5 @@
     public static final String MESSAGE_ADDING_DOCKER_SUPPORT = "Adding docker support ...";
     public static final String MESSAGE_DOCKER_HOST_INFO = "Current docker host: %s";
     public static final String MESSAGE_EXECUTE_DOCKER_RUN = "Executing Docker Run...";
+    public static final String ERROR_BUILDING_IMAGE = "Fail to build image. Please verify docker host connection and Dockerfile.";
```
adashen: Currently the dock host is localhost right? So can we have more specific message?

Issue #891, opened 2017-08-14
[Intellij] It cannot throw the exception when debug


ruoxinCheng:
eg: the output file is already exist, it cannot show it in console panel.

ruoxinCheng:
Verified it.

PR #900, opened 2017-08-14
Fix the log missing issue of Spark Job containers

wezhang:
Replace the container log fetching stop from unsubscribe()
to a dedicated Observable since those logs need to handle
before being unsubscribed.


wezhang:
to fix #891

PR #913, opened 2017-08-15
upgrade version to 3.0.7

adashen:


PluginsAndFeatures/azure-toolkit-for-intellij/resources/META-INF/plugin.xml
```diff
@@ -11,10 +11,10 @@
   <change-notes>
     <![CDATA[
     <html>
-      <h3>[3.0.6]</h3>
+      <h3>[3.0.7]</h3>
```
jdneo: reserve the change notes history

Issue #915, opened 2017-08-15
[Eclipse] Cannot clicking Confirm button to install scala IDE


ruoxinCheng:
1.  open eclipse
2.  Cancel the scala installation.
3.  Create a project
4,  popup an dialog to click on to install
Actual:
Cannot click 'confirm' button to install

djlan:
Fixed

ruoxinCheng:
Verified it.

PR #916, opened 2017-08-15
Fix bug where the marketplace window cannot get focus due to model wi…

djlan:
…ndow created


djlan:
Bug https://github.com/Microsoft/azure-tools-for-java/issues/915

Issue #992, opened 2017-09-05
[IntelliJ] When change the different cluster, the password and username always passed


ruoxinCheng:

ruoxinCheng:
Yes, verified it.

PR #994, opened 2017-09-05
Fix the issue of selected cluster not apply to SSH auth verify

wezhang:


wezhang:
to fix #992

Issue #796, opened 2017-07-27
[IntelliJ]  The SSH IDE always keep the last setting


ruoxinCheng:
1.  Delete original debug configurations.
2.  Add a new configuration
Actual:  SSH keeps the last setting
Expect:  SSH keeps the empty state.

1. Create one debug configurations and set SSH IDE
2.  Create another debug configuration and set SSH IDE
Actual:  SSH keeps the last setting
Expect:  SSH keeps the empty state.

ruoxinCheng:
Verified it.

PR #995, opened 2017-09-05
Fix the issue that SSH not empty for new run configuration

wezhang:


wezhang:
to fix #796

PR #1002, opened 2017-09-07
select/deselect all subscriptions

Eskibear:


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azuretools/ijidea/ui/SubscriptionsDialog.java
```diff
@@ -207,4 +226,20 @@ protected String getDimensionServiceKey() {
         return "SubscriptionsDialog";
     }
 
+    private static class SubscriptionTableModel extends DefaultTableModel {
```
jdneo: Need to be static here?

PR #1012, opened 2017-09-11
Add select docker file filed in docker run

jdneo:
This PR allows user to select a docker file when he wants to use the **Docker Run** feature. So no need to run **Add Docker Support** first.

UI:
It will load the default docker file under the project base path. (If the user has already clicked the **Add Docker Support** before).
image

If the docker file is not specified or the file is not existed, a error message will show at the bottom of the RunDialog.

image

image

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/container/dockerhost/DockerHostRunState.java
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/container/dockerhost/DockerHostRunState.java
@@ -1,3 +1,25 @@
+/*
+ * Copyright (c) Microsoft Corporation
+ *
+ * All rights reserved.
+ *
+ * MIT License
+ *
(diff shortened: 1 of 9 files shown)


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/container/dockerhost/DockerHostRunState.java
```diff
@@ -98,7 +120,7 @@ public void onTextAvailable(ProcessEvent processEvent, Key key) {
                     processHandler.setText(String.format("Locating artifact ... [%s]", targetFilePath));
 
                     // validate dockerfile
-                    Path targetDockerfile = Paths.get(basePath, Constant.DOCKERFILE_FOLDER, Constant.DOCKERFILE_NAME);
+                    Path targetDockerfile = Paths.get(dataModel.getDockerFilePath());
```
adashen: error handling ?

Issue #1014, opened 2017-09-11
"The request content was invalid and could not be deserialized" when using Run on Web App (Linux)


jdneo:
**Repro step:**
1. Create a web app on linux on Azure Portal
![image](https://user-images.githubusercontent.com/6193897/30266026-9e315de2-9710-11e7-9ce0-9b25d6a3e882.png)

2. In IntelliJ, Using **Run on Web App (Linux)** for the created web app (Linux)
![image](https://user-images.githubusercontent.com/6193897/30266267-5dfd642c-9711-11e7-93ca-fa567dc42ddc.png)

3. Click Run

**Expect:**
The web app (Linux) will be updated

**Result:** 
An error is throwed.

**Error Message:**
The request content was invalid and could not be deserialized: 'Error reading string. Unexpected token: StartObject. Path 'tags.hidden-related:/subscriptions/<subscription-id>/resourcegroups/test-app-api/providers/Microsoft', line 1, position 152.'.

@bsaby 
Please help double check if this issue do exist in our current released plugin, thanks.

bsaby:
Also repro in our current released plugin for Eclispe and IntelliJ

**Eclipse:**
![image](https://user-images.githubusercontent.com/22470312/30306226-9f183d54-97a9-11e7-9409-aec0e2b323da.png)

**IntelliJ:**
![image](https://user-images.githubusercontent.com/22470312/30306304-f9a96bda-97a9-11e7-9994-5ee05caf4ef6.png)

jdneo:
@bsaby Fixed, please help verify this in the new build, thanks.

bsaby:
Test passed.
Build: dev383

PR #1018, opened 2017-09-12
fix tag error

jdneo:
Related Issue: #1014 

Root cause is the tag messed up in the request body. [See the Issue here](https://github.com/Azure/azure-sdk-for-java/issues/1755).

So the work arround is removing the tag before update.

Here is the [reference](https://github.com/Microsoft/azure-maven-plugins/blob/master/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/WebAppUtils.java#L73).


Utils/azuretools-core/src/com/microsoft/azuretools/core/mvp/model/webapp/AzureWebAppMvpModel.java
```diff
@@ -389,4 +390,15 @@ public void cleanWebAppsOnLinux() {
...
     }
+
+    /**
+     * Work Around:
+     * When a web app is created from Azure Portal, there are hidden tags associated with the app.
+     * It will be messed up when calling "update" API.
+     * An issue is logged at https://github.com/Azure/azure-sdk-for-java/issues/1755 .
+     * Remove all tags here to make it work.
+     */
+    private void clearTags(final WebApp app) {
```
adashen: notnull?

PR #1039, opened 2017-09-20
refactor auth flow

adashen:
Don't merge!
Modify auth flow


Utils/azuretools-core/src/com/microsoft/azuretools/adauth/CacheDriver.java
```diff
@@ -0,0 +1,166 @@
+package com.microsoft.azuretools.adauth;
```
jdneo: lisence

PR #1044, opened 2017-09-21
extract common UI

jdneo:
This PR extract the common UI part in container and enabled ACR selector in push image feature.

UI change:
Loading
image

After loading
image

Select ACR
image

not select ACR
image

### DIFF
--- a/Utils/azuretools-core/src/com/microsoft/azuretools/core/mvp/model/container/ContainerRegistryMvpModel.java
+++ b/Utils/azuretools-core/src/com/microsoft/azuretools/core/mvp/model/container/ContainerRegistryMvpModel.java
@@ -92,7 +92,7 @@ public class ContainerRegistryMvpModel {
     /**
      * Get Registry Credential.
      */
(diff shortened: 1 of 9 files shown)


Utils/azuretools-core/src/com/microsoft/azuretools/core/mvp/model/container/ContainerRegistryMvpModel.java
```diff
@@ -49,28 +60,56 @@ public static ContainerRegistryMvpModel getInstance() {
...
+    }
+
+    /**
+     * Get Registry Credential.
+     */
+    public PrivateRegistryImageSetting getRegistryCredential(@NotNull final Registry registry) throws Exception {
+        if (!registry.adminUserEnabled()) {
+            throw new Exception(ADMIN_USER_NOT_ENABLED);
+        }
+        final RegistryListCredentials credentials = registry.listCredentials();
```
adashen: will listCredentials() throw exception? If so, catch and wrap the exception

PR #1077, opened 2017-09-28
caught auth required exception in adauthmanager and ask user to sign …

adashen:
…in again


Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/AdAuthManager.java
```diff
@@ -76,7 +80,15 @@ public String getAccessToken(String tid) throws IOException {
      */
     public String getAccessToken(String tid, String resource, PromptBehavior promptBehavior) throws IOException {
         AuthContext ac = createContext(tid, null);
-        AuthResult result = ac.acquireToken(resource, false, adAuthDetails.getAccountEmail(), false);
+        AuthResult result = null;
+        try {
+            result = ac.acquireToken(resource, false, adAuthDetails.getAccountEmail(), false);
+        } catch (AuthException e) {
+            if (AuthError.InvalidGrant.equalsIgnoreCase(e.getError())
```
jdneo: we can throw the exception if not going into the if block.

Issue #1081, opened 2017-09-28
Image and tag field is not right


jdneo:
If the user restart the IntelliJ and open push image panel (or web app for linux), the imge and tag field will contain server url.

<img width="618" alt="screen shot 2017-09-28 at 9 10 07 pm" src="https://user-images.githubusercontent.com/6193897/30968305-7f331392-a424-11e7-9035-1624c13ff56a.png">

jdneo:
@bsaby fixed, please help verify, thanks.

bsaby:
Test passed.
BUild: Dev418

PR #1087, opened 2017-09-30
bug fix

jdneo:
Related issue: #1081 

Add a new method in ```PrivateRegistryImageSetting```

The new method ```getImageTagWithServerUrl()``` is used to get the combination of server URL and image with tag

### DIFF
--- a/Utils/azuretools-core/src/com/microsoft/azuretools/core/mvp/model/webapp/PrivateRegistryImageSetting.java
+++ b/Utils/azuretools-core/src/com/microsoft/azuretools/core/mvp/model/webapp/PrivateRegistryImageSetting.java
@@ -69,11 +69,10 @@ public class PrivateRegistryImageSetting extends ImageSetting {
         this.password = password;
     }
 
-    @Override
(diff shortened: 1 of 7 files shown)


Utils/azuretools-core/src/com/microsoft/azuretools/core/mvp/model/webapp/PrivateRegistryImageSetting.java
```diff
@@ -69,11 +69,10 @@ public void setPassword(String password) {
         this.password = password;
     }
 
-    @Override
-    public String getImageNameWithTag() {
+    public String getImageTagWithServerUrl() {
         if (this.serverUrl != null && super.getImageNameWithTag() != null) {
```
Eskibear: better to make member `protected` in super class

jdneo: will fix this in next PR

PR #1095, opened 2017-10-09
bug fix

jdneo:
- Fix an interactive login bug
- Make the error message more specific when cannot get azure manager


Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/AuthMethodManager.java
```diff
@@ -48,7 +48,8 @@
 
 public class AuthMethodManager {
     private final static Logger LOGGER = Logger.getLogger(AuthMethodManager.class.getName());
-    private static final String CANNOT_GET_AZURE_MANAGER = "Cannot get Azure Manager.";
+    private static final String CANNOT_GET_AZURE_MANAGER = "Cannot get Azure Manager with Subscription ID: %s. "
+            + "Please check if you have already signed in.";
```
adashen: shall we ask user to sign in with this subscription?

PR #1097, opened 2017-10-09
Update READ.md with HDInsight Links

Matthew-Dong:
Update READ.md with HDInsight Links


vscjavaci:

Issue #1109, opened 2017-10-12
[Eclipse] Doesn't display subscription after login via Azure toolkit


jacwu:
Follow https://docs.azure.cn/zh-cn/articles/others/developerdifferences regarding to how to use Java in Eclipse and connect to China Azure, after login, Eclipse doesn't show any subscription.
![capture](https://user-images.githubusercontent.com/16080993/31484131-a1ca7f9c-af61-11e7-8759-9ca01abcfab1.JPG)

Eskibear:
Azure SDK issue, https://github.com/Azure/autorest-clientruntime-for-java/issues/263

By reverting to version 1.0.4, it works:
![image](https://user-images.githubusercontent.com/2351748/31528223-c14dead6-b004-11e7-844a-4a1e430109b9.png)

adashen:
@bsaby we have updated the sdk. And the issue should be fixed. Please verify it on the latest develop build

Eskibear:
fixed in #1145

PR #1114, opened 2017-10-13
[ONHOLD] workaround for mooncake login issue

Eskibear:
related issue: https://github.com/Microsoft/azure-tools-for-java/issues/1109

root cause: https://github.com/Azure/autorest-clientruntime-for-java/issues/263


Eskibear:
ONHOLD. potential dependency compatibility problem.

Issue #1028, opened 2017-09-14
[Eclispe][Mac]Redis Cache-> Eclipse will hang up when creating an existing redis


bsaby:
1. Input an existing redis name
2. Click OK

Expect:
The Alert dialog box pops up
![image](https://user-images.githubusercontent.com/22470312/30412618-54216336-994b-11e7-8c2c-5910edc8c122.png)
Result:
The create dialog box will stop responding
![image](https://user-images.githubusercontent.com/22470312/30412578-09605ec4-994b-11e7-8853-16f191c34fb0.png)

jdneo:
@bsaby Fixed, please help verify, thanks.

songkaiape:
Fixed.

bsaby:
Build used: dev440

PR #1118, opened 2017-10-13
Fix eclipse redis create bug

jdneo:
Related Issue: #1028


Utils/azuretools-core/src/com/microsoft/azuretools/azurecommons/exceptions/InvalidFormDataException.java
```diff
@@ -0,0 +1,29 @@
...
+ * THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO
+ * THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package com.microsoft.azuretools.azurecommons.exceptions;
+
+public class InvalidFormDataException extends Exception {
```
adashen: I remember there is similar class in the containerregistry composite right?

jdneo: Yes, and I think it will be better if we move it into core, because a lot of place will use it.

Issue #1117, opened 2017-10-13
[Eclipse]Update the plugin without updating the web App on Linux name to web app for containers


bsaby:
1. First Install a build number < dev432 (the previous build number we used is dev428)
2. Uninstall old build or update new  build directly, new build number>= dev432

The name of web app on linux does not become to web app for containers

![image](https://user-images.githubusercontent.com/22470312/31531461-d8826f5a-b019-11e7-94d5-4e89899cfcfd.png)

adashen:
fixed, please verified on latest developer build (440 or later).

bsaby:
Test passed. 
Build: dev440
The name will be updated after updating the plugin.

PR #1120, opened 2017-10-13
fix rename issue #1117

adashen:
It seems this is a eclipse issue, Rename an already registered command, when updating the plugin in eclipse, the name stays the same.

So the workaround in this fix, is use a new command id and command name.
@bsaby since the command id has been modified. So please verify the whole feature of publish to web app for containers to make sure there is no regression.


bsaby:
Have verified.
No regression issue.

Issue #1126, opened 2017-10-16
[IntelliJ]Can't save default run configuration


bsaby:
1. Open run configuration
2. Expand Defaults->Azure Docker Support
3. Select any sub item under Azure Docker Support
4. Edit configuration and save them
![image](https://user-images.githubusercontent.com/22470312/31601805-ab01d8ec-b28d-11e7-8ebb-2c62b6b3b048.png)

**On Mac/Linux**
An error dialog pops up when clicking  Apply or OK 
![image](https://user-images.githubusercontent.com/22470312/31602213-ec12b850-b28e-11e7-9f74-87661089dc8a.png)

Sometimes, pops up below error
![image](https://user-images.githubusercontent.com/22470312/31601847-c7601328-b28d-11e7-8199-451c7c9cb2f6.png)

On Windows: 
No error after clicking OK or Apply, but when click "+" on Upper left corner to add the same item, the default configuration will not displayed

jdneo:
Seems we need to remove the overrided ```createConfiguration``` in ```*ConfigurationFactory```

Eskibear:
For the error dialogs, I think it has nothing to do with platform. I suppose that's because you have clicked `Web App for Containers` before. When clicking `OK`, it will try to save both configurations  , and fail the validation of  `Web App for Containers`. @bsaby

bsaby:
Yes, try this step again, the error dialog pops up because click Web App for Containers before, it is nothing to do with OS.

adashen:
low priority.

Eskibear:
It's determined by IntelliJ, which validate every configuration you have ever opened in current window.

PR #1131, opened 2017-10-17
fix bug: load default run configuration at creation

Eskibear:
#1126


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/container/pushimage/PushImageRunConfiguration.java
```diff
@@ -87,6 +87,13 @@ public PushImageRunModel getDataModel() {
     }
 
     @Override
+    public RunConfiguration clone() {
```
adashen: firstTimeCreated value will also be cloned?

adashen:
will cause new created configuration based on default without before run task.

PR #1149, opened 2017-10-23
add docker file selector in ACR composite

Eskibear:


PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.container/src/main/java/com/microsoft/azuretools/container/ui/common/ContainerSettingComposite.java
```diff
@@ -81,31 +84,40 @@ public ContainerSettingComposite(Composite parent, int style, String basePath) {
         projectBasePath = basePath;
         registryCache = new ArrayList<>();
 
-        setLayout(new GridLayout(3, false));
+        setLayout(new GridLayout(4, false));
+
+        lblDockerFile = new Label(this, SWT.NONE);
+        lblDockerFile.setText("Docker File");
```
adashen: const

PR #1182, opened 2017-10-26
fix listing HDInsight cluster error in Mooncake environment

### LINKED ISSUE #1173: list HDInsight clusters error in MoonCake environment
Cannot list HDInsight clusters in Azure China environment.

voidfunction:
Fix #1173 list HDInsight clusters error in MoonCake environment

### DIFF
--- a/Utils/azuretools-core/src/com/microsoft/azuretools/sdkmanage/ServicePrincipalAzureManager.java
+++ b/Utils/azuretools-core/src/com/microsoft/azuretools/sdkmanage/ServicePrincipalAzureManager.java
@@ -170,7 +170,21 @@ public class ServicePrincipalAzureManager extends AzureManagerBase {
     public String getManagementURI() throws IOException {
         initATCIfNeeded();
         // default to global cloud
-        return atc.environment() == null ? AzureEnvironment.AZURE.resourceManagerEndpoint() : atc.environment().resourceManagerEndpoint();
+        String managementUri = atc.environment() == null ? AzureEnvironment.AZURE.resourceManagerEndpoint() : atc.environment().resourceManagerEndpoint();
+
+        /*
+        this is a walking around for MoonCake resource manager endpoint
+        It should be 'https://management.chinacloudapi.cn/', but Mooncake environment return 'https://management.chinacloudapi.cn'
+        The following error message can be expected without the walking around
+
+        {
+        "error": {
+            "code": "InvalidAuthenticationTokenAudience",
+            "message": "The access token has been obtained from wrong audience or resource 'https://management.chinacloudapi.cn'. It should exactly match (including forward slash) with one of the allowed audiences 'https://management.core.chinacloudapi.cn/','https://management.chinacloudapi.cn/'."
+            }
+        }
+         */
+        return managementUri.endsWith("/") ? managementUri : managementUri + "/";
     }
 
     @Override
--- a/Utils/azuretools-core/src/com/microsoft/azuretools/sdkmanage/ServicePrincipalAzureManager.java
+++ b/Utils/azuretools-core/src/com/microsoft/azuretools/sdkmanage/ServicePrincipalAzureManager.java
@@ -170,21 +170,7 @@ public class ServicePrincipalAzureManager extends AzureManagerBase {
     public String getManagementURI() throws IOException {
         initATCIfNeeded();
         // default to global cloud
-        String managementUri = atc.environment() == null ? AzureEnvironment.AZURE.resourceManagerEndpoint() : atc.environment().resourceManagerEndpoint();
-
-        /*
-        this is a walking around for MoonCake resource manager endpoint
-        It should be 'https://management.chinacloudapi.cn/', but Mooncake environment return 'https://management.chinacloudapi.cn'
-        The following error message can be expected without the walking around
-
-        {
-        "error": {
-            "code": "InvalidAuthenticationTokenAudience",
-            "message": "The access token has been obtained from wrong audience or resource 'https://management.chinacloudapi.cn'. It should exactly match (including forward slash) with one of the allowed audiences 'https://management.core.chinacloudapi.cn/','https://management.chinacloudapi.cn/'."
-            }
-        }
-         */
-        return managementUri.endsWith("/") ? managementUri : managementUri + "/";
+        return atc.environment() == null ? AzureEnvironment.AZURE.resourceManagerEndpoint() : atc.environment().resourceManagerEndpoint();
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


vscjavaci:

Utils/azuretools-core/src/com/microsoft/azuretools/sdkmanage/ServicePrincipalAzureManager.java
```diff
@@ -170,7 +170,21 @@ public String getAccessToken(String tid) throws IOException {
...
+        The following error message can be expected without the walking around
+
+        {
+        "error": {
+            "code": "InvalidAuthenticationTokenAudience",
+            "message": "The access token has been obtained from wrong audience or resource 'https://management.chinacloudapi.cn'. It should exactly match (including forward slash) with one of the allowed audiences 'https://management.core.chinacloudapi.cn/','https://management.chinacloudapi.cn/'."
+            }
+        }
+         */
+        return managementUri.endsWith("/") ? managementUri : managementUri + "/";
```
adashen: should not change the endpoint. it is from azure environment

djlan: To clarify, it doesn't change the endpoint, just adds "/" at the end.

---
In reply to: [147088515](https://github.com/Microsoft/azure-tools-for-java/pull/1182#discussion_r147088515) [](ancestors = 147088515)

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/cluster/ClusterOperationImpl.java
```diff
@@ -51,25 +50,23 @@
...
-                                 new RestServiceManagerBaseImpl());
-                    }
-               });
+              String response = requestWithToken(subscription.getTenantId(), (accessToken) -> {
+                      String managementUrl = AuthMethodManager.getInstance().getAzureManager().getManagementURI();
+                      return AzureAADHelper.executeRequest(managementUrl,
+                              String.format("/subscriptions/%s/providers/Microsoft.HDInsight/clusters?api-version=%s", subscription.getSubscriptionId(), VERSION),
+                              RestServiceManager.ContentType.Json,
+                              "GET",
+                              null,accessToken,
```
djlan: >null,accessToken, [](start = 30, length = 17)

Nit, two lines.

Utils/azuretools-core/src/com/microsoft/azuretools/sdkmanage/ServicePrincipalAzureManager.java
voidfunction: `AzureEnvironment.AZURE_CHINA.resourceManagerEndpoint()` return `https://management.chinacloudapi.cn/`.
But `atc.environment().resourceManagerEndpoint()` return `https://management.chinacloudapi.cn` in Mooncake(missing /). Which makes an error when querying rest api.
``` {
      "error": {
            "code": "InvalidAuthenticationTokenAudience",
            "message": "The access token has been obtained from wrong audience or resource 'https://management.chinacloudapi.cn'. It should exactly match (including forward slash) with one of the allowed audiences 'https://management.core.chinacloudapi.cn/','https://management.chinacloudapi.cn/'."
            }
        } 
```


Actually I just found `AzureEnvironment.AZURE_CHINA.resourceManagerEndpoint()` return the right endpoint, but `atc.environment().resourceManagerEndpoint()` not.
image

adashen: the atc environment didn't return / is because of the endpoint in user authfile doen't has that. This could be caused by the different versions of code to generate the authfile. However, we would like to make this method just return the url as it is. How about do your check in your own code logic? And it also could be the case AZURE_CHINA.resouceManagerEndpoint will remove the tailing slash. We just experienced an issue in this sprint, what caused by the slash change in sdk.

Besides, to get the manageURI, please use AuthMethodManager.getInstance().getAzureManager().getEnvironment() to get the AzureEnvironment first. Then you could use the environment to get all the endpoints

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/cluster/ClusterOperationImpl.java
```diff
@@ -51,25 +50,23 @@
...
-                                 String.format("api/Clusters/GetAll?subscriptionIds=%s;&_=%d", subscription.getSubscriptionId(), new Date().getTime()),
-                                 RestServiceManager.ContentType.Json,
-                                 "GET",
-                                 null,
-                                 accessToken,
-                                 new RestServiceManagerBaseImpl());
-                    }
-               });
+              String response = requestWithToken(subscription.getTenantId(), (accessToken) -> {
+                      String managementUrl = AuthMethodManager.getInstance().getAzureManager().getManagementURI();
```
adashen: Instead of getManagementURi like this would suggest to use AuthMethodManager.getInstance().getAzureManager().getEnvironment() first to get the azure environment.

Utils/azuretools-core/src/com/microsoft/azuretools/sdkmanage/ServicePrincipalAzureManager.java
voidfunction: The auth file is generated by azure cli. `AZURE_CHINA.resouceManagerEndpoint` do have the slash. I think we should cover this case if customers auth file is generated by CLI too.

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/cluster/ClusterOperationImpl.java
voidfunction: Fixed. Thanks.

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/cluster/ClusterOperationImpl.java
voidfunction: Fixed. Thanks.

Utils/azuretools-core/src/com/microsoft/azuretools/sdkmanage/ServicePrincipalAzureManager.java
voidfunction: Remove this file change since we have another way to get the endpoint. Thanks, Wei.

Issue #1215, opened 2017-11-02
[IntelliJ] Remove Label "WINUTILS.exe" on Mac and Linux in Configuration page.


tankungx:

tankungx:
verified.

PR #1228, opened 2017-11-06
Hide the Winutil location label for non-windows platforms

wezhang:


wezhang:
will fix #1215

Issue #1240, opened 2017-11-07
[IntelliJ] The context menu only work on classname or beyond script


ruoxinCheng:
![image](https://user-images.githubusercontent.com/20041888/32484899-60bc3632-c3dc-11e7-8b5b-7e23e8831b2c.png)
![image](https://user-images.githubusercontent.com/20041888/32484928-7ab35750-c3dc-11e7-9e68-3f115d740eba.png)

wezhang:
Only repro on Scala file job.

wezhang:
The Local Run Configuration sorted as the 2nd one.

ruoxinCheng:
Verified it.

Issue #1211, opened 2017-11-02
[IntelliJ] These options are confused from context menu


ruoxinCheng:
Right-click from script editor and from project explorer.
![image](https://user-images.githubusercontent.com/20041888/32308133-34e02922-bfbf-11e7-83c4-61d1ffb93c51.png)

wezhang:
I'll try to remove them.

wezhang:
Run context feature added

ruoxinCheng:
It work on 2017.2 , but not work on 2017.1.

ruoxinCheng:
and SBT also not support context menu.

wezhang:
Due to 2017.1 Scala plugin limitation, Local Run Context only works in at `File`, `Object`, `main` function of scala file. No problems in java file.

ruoxinCheng:
Verified it.

PR #1242, opened 2017-11-08
Fix Local Run Context not recognized in code issue

wezhang:
Don't update the source PsiElement to avoid common application
run configuration getting higher priority.


wezhang:
would fix #1240

wezhang:
Added fix for #1211, not working at sbt project issue.

PR #1255, opened 2017-11-17
 Fix bug where link in message dialog is not clickable

djlan:


vscjavaci:

Issue #1254, opened 2017-11-16
[IntelliJ][HDInsight] Got NPE sometimes


wezhang:
```
update failed for AnAction with ID=RunClass
java.lang.NullPointerException
	at com.microsoft.azure.hdinsight.spark.run.SparkBatchJobLocalRunConfigurationProducer.lambda$setupConfigurationFromContext$2(SparkBatchJobLocalRunConfigurationProducer.java:71)
	at java.util.Optional.filter(Optional.java:178)
	at com.microsoft.azure.hdinsight.spark.run.SparkBatchJobLocalRunConfigurationProducer.setupConfigurationFromContext(SparkBatchJobLocalRunConfigurationProducer.java:68)
	at com.microsoft.azure.hdinsight.spark.run.SparkBatchJobLocalRunConfigurationProducer.setupConfigurationFromContext(SparkBatchJobLocalRunConfigurationProducer.java:60)
	at com.intellij.execution.actions.RunConfigurationProducer.createConfigurationFromContext(RunConfigurationProducer.java:90)
	at com.intellij.execution.junit.JavaRunConfigurationProducerBase.createConfigurationFromContext(JavaRunConfigurationProducerBase.java:64)
	at com.intellij.execution.actions.RunConfigurationProducer.findOrCreateConfigurationFromContext(RunConfigurationProducer.java:178)
	at com.intellij.execution.actions.PreferredProducerFind.getConfigurationsFromContext(PreferredProducerFind.java:116)
	at com.intellij.execution.actions.PreferredProducerFind.findConfigurationFromContext(PreferredProducerFind.java:141)
	at com.intellij.execution.actions.PreferredProducerFind.createConfiguration(PreferredProducerFind.java:44)
	at com.intellij.execution.actions.ConfigurationContext.createConfiguration(ConfigurationContext.java:148)
	at com.intellij.execution.actions.ConfigurationContext.getConfiguration(ConfigurationContext.java:139)
	at com.intellij.execution.actions.BaseRunConfigurationAction.update(BaseRunConfigurationAction.java:187)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.performDumbAwareUpdate(ActionUtil.java:132)
	at com.intellij.openapi.actionSystem.impl.Utils.doUpdate(Utils.java:235)
	at com.intellij.openapi.actionSystem.impl.Utils.expandActionGroup(Utils.java:180)
[...]

PR #1257, opened 2017-11-22
Fix the nullable module caused NPE issue in context run

wezhang:


wezhang:
would fix #1254

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/spark/run/SparkBatchJobLocalRunConfigurationProducer.java
```diff
@@ -91,6 +77,25 @@ protected boolean setupConfigurationFromContext(RemoteDebugRunConfiguration conf
                 .orElse(false);
```
voidfunction: I agree that Optional makes our code robust. But I think the style is not very friendly for debugging.

wezhang: You're right, fortunately, IDE can help some Lambda function breakpoints, we can use multi-lines style to meet IDE requirements.

Issue #1258, opened 2017-11-22
[IntelliJ][HDInsight] GetClusterConfiguration error since ContentType is null


wezhang:
```
     public ClusterConfiguration getClusterConfiguration(final SubscriptionDetail subscription, final String clusterId) throws IOException, HDIException, AzureCmdException {
          try {
               String response = requestWithToken(subscription.getTenantId(), new RequestCallback<String>() {
                    @Override
                    public String execute(String accessToken) throws Throwable {
                         String managementURI = AuthMethodManager.getInstance().getAzureManager().getManagementURI();
                         return AzureAADHelper.executeRequest(managementURI,
                                 String.format("%s/configurations?api-version=%s", clusterId.replaceAll("/+$", ""), VERSION),
                                 null,
                                 "GET",
                                 null,
                                 accessToken,
                                 new RestServiceManagerBaseImpl());
                    }
               });
```

wezhang:
Should pass `ContentType.Json` as `executeRequest` 2nd argument

ruoxinCheng:
Maybe same as #1266

ruoxinCheng:
Maybe same as #1266

ruoxinCheng:
Verified.

Issue #1253, opened 2017-11-16
[IntelliJ] Disconnect button should be disabled when the job have not submit


ruoxinCheng:
![image](https://user-images.githubusercontent.com/20041888/32880165-9b44e292-cae7-11e7-9a8e-3a3fc106b4e8.png)

ruoxinCheng:
Verified.

PR #1259, opened 2017-11-22
Fix the Spark Job Run disconnect button issue

wezhang:
The button should be enabled after job is submitted.


wezhang:
would fix #1258, #1253

Issue #1244, opened 2017-11-09
[intellij] [hdinsight]Spark Job Remote Run log console doesn’t handle well for retried job


wezhang:

ruoxinCheng:
Verified.

PR #1260, opened 2017-11-23
Fix the Spark Job Submission can't get retry log issue

wezhang:


wezhang:
would fix #1244

Issue #1210, opened 2017-11-02
[IntelliJ] Add output path like 'wasb://123@blob1/HVACout2missed', It comeout from __default__ folder


ruoxinCheng:
If exists HVACout2missed, it says exist folder.

ruoxinCheng:
The first folder '123@blob1' cannot created. and it show 
java.lang.IllegalArgumentException: Can not create a Path from an empty string
	at org.apache.hadoop.fs.Path.checkPathArg(Path.java:127)
	at org.apache.hadoop.fs.Path.<init>(Path.java:135)
	at org.apache.hadoop.fs.Path.<init>(Path.java:94)
	at com.microsoft.azure.hdinsight.spark.mock.MockRawLocalFileSystem.pathToFile(MockRawLocalFileSystem.java:110)
	at org.apache.hadoop.fs.RawLocalFileSystem.mkdirs(RawLocalFileSystem.java:421)
	at org.apache.hadoop.fs.RawLocalFileSystem.mkdirs(RawLocalFileSystem.java:427)
	at org.apache.hadoop.fs.RawLocalFileSystem.mkdirs(RawLocalFileSystem.java:427)
	at org.apache.hadoop.fs.RawLocalFileSystem.mkdirs(RawLocalFileSystem.java:427)
	at org.apache.hadoop.fs.ChecksumFileSystem.mkdirs(ChecksumFileSystem.java:584)
	at org.apache.hadoop.mapreduce.lib.output.FileOutputCommitter.setupJob(FileOutputCommitter.java:291)
	at org.apache.hadoop.mapred.FileOutputCommitter.setupJob(FileOutputCommitter.java:131)
	at org.apache.spark.SparkHadoopWriter.preSetup(SparkHadoopWriter.scala:64)
	at org.apache.spark.rdd.PairRDDFunctions$$anonfun$saveAsHadoopDataset$1.apply$mcV$sp(PairRDDFunctions.scala:1188)
	at org.apache.spark.rdd.PairRDDFunctions$$anonfun$saveAsHadoopDataset$1.apply(PairRDDFunctions.scala:1161)
	at org.apache.spark.rdd.PairRDDFunctions$$anonfun$saveAsHadoopDataset$1.apply(PairRDDFunctions.scala:1161)
	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)
[...]

PR #1268, opened 2017-11-28
Fix Mocked fs parse non-default container root path issue

wezhang:
Also introduced Windows file status class with fixing reference Class
RawLocalFileSystem method loadPermissionInfo() issue.


wezhang:
To fix #1210

PR #1277, opened 2017-11-29
Make DFS mock compatible with Hadoop common 2.2

wezhang:


wezhang:
Would fix #1297

wezhang:
Would fix #1295

Utils/pom.xml
```diff
@@ -257,7 +257,7 @@
             <dependency>
                 <groupId>org.apache.hadoop</groupId>
                 <artifactId>hadoop-common</artifactId>
-                <version>2.7.1</version>
+                <version>2.2.0</version>
```
voidfunction: ><version>2.2.0</version> [](start = 16, length = 24)

down to 2.2?

wezhang:
Would fix #1298 

image

Utils/pom.xml
wezhang: has to... since the current spark-core 2.0/2.1 dependent on 2.2.0

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/spark/run/SparkBatchJobDebuggerRunner.java
```diff
@@ -120,11 +120,28 @@ public GenericDebuggerRunnerSettings createConfigurationData(ConfigurationInfoPr
         return null;
     }
 
+    private void checkSubmissionParameter(SparkSubmissionParameter parameter) throws ExecutionException {
+        if (StringUtils.isEmpty(parameter.getArtifactName())) {
+            throw new ExecutionException("The artifact to submit is not selected, please config it at Run/Debug configuration -> Remotely Run in Cluster.");
```
djlan: >throw new ExecutionException("The artifact to submit is not selected, please config it at Run/Debug configuration -> Remotely Run in Cluster."); [](start = 12, length = 144)

Question: This will show up in the user UI right? Because this will need user to take some action.

PR #1310, opened 2017-12-25
modify UA according to the telemetry setting

jdneo:
TODO: Eclipse side will be in another PR


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/ui/AzurePanel.java
```diff
@@ -101,6 +103,11 @@ public boolean doOKAction() {
                     // Either from Agree to Deny, or from Deny to Agree.
                     final String action = acceptTelemetry ? AppInsightsConstants.Allow : AppInsightsConstants.Deny;
                     AppInsightsClient.createByType(AppInsightsClient.EventType.Telemetry, "", action, null, true);
+                    String userAgent = (String.format(AzurePlugin.USER_AGENT, CommonConst.PLUGIN_VERISON));
+                    if (acceptTelemetry) {
+                        userAgent += String.format(", machineid:%s", DataOperations.getProperty(dataFile, message("instID")));
```
Eskibear: what if `DataOperations.getProperty(dataFile, message("instID"))` is empty string? just leave it until next restart?

PR #1311, opened 2017-12-26
modify UA according to the telemetry setting

jdneo:
This is for Eclipse


PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/Activator.java
```diff
@@ -115,15 +117,20 @@ public Activator() {
      * (non-Javadoc)
      * @see org.eclipse.ui.plugin.AbstractUIPlugin#start(org.osgi.framework.BundleContext)
      */
+    @Override
     public void start(BundleContext context) throws Exception {
         super.start(context);
         plugin = this;
+        pluginInstLoc = String.format("%s%s%s", PluginUtil.pluginFolder, File.separator,
```
Eskibear: path.join?

PR #1340, opened 2018-01-12
Fix the issues when linking a secure hadoop cluster

wezhang:

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/serverexplore/ui/AddNewClusterFrom.java
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/serverexplore/ui/AddNewClusterFrom.java
@@ -21,10 +21,13 @@
  */
 package com.microsoft.azure.hdinsight.serverexplore.ui;
 
+import com.intellij.CommonBundle;
+import com.intellij.ide.BrowserUtil;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.application.ModalityState;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.ui.DialogWrapper;
+import com.intellij.openapi.ui.popup.IconButton;
 import com.microsoft.azure.hdinsight.common.ClusterManagerEx;
 import com.microsoft.azure.hdinsight.sdk.cluster.HDInsightAdditionalClusterDetail;
 import com.microsoft.azure.hdinsight.sdk.common.HDIException;
@@ -41,7 +44,7 @@ import org.jetbrains.annotations.Nullable;
 import javax.swing.*;
 import java.awt.*;
 import java.awt.event.ActionEvent;
-import java.awt.event.ActionListener;
+import java.util.stream.Stream;
 
 public class AddNewClusterFrom extends DialogWrapper {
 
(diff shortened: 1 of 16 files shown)


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/serverexplore/ui/AddNewClusterFrom.java
```diff
@@ -63,18 +66,23 @@
...
+    private JLabel passwordLabel;
 
     private HDInsightRootModule hdInsightModule;
 
     private static final String URL_PREFIX = "https://";
+    private static final String HELP_URL = "https://go.microsoft.com/fwlink/?linkid=866472";
 
     public AddNewClusterFrom(final Project project, HDInsightRootModule hdInsightModule) {
         super(project, true);
+        myHelpAction = new HelpAction();
```
djlan: >myHelpAction = new HelpAction(); [](start = 8, length = 32)

1. Can we add a "HyperLink" in the dialogue so users can click/know if they run into issues?
2. DialogWrapper seems already have HelpAction defined, no need to add that part of code here?

djlan: Or add tooltip

---
In reply to: [161944649](https://github.com/Microsoft/azure-tools-for-java/pull/1340#discussion_r161944649) [](ancestors = 161944649)

wezhang: Currently, the Intellij style 'help' icon is help to open helplink. The HelpAction defined by DialogWrapper is only used for Intellij internal help DB with ID reference. Fix it by adding more details into tooltip.

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/ClusterManagerEx.java
```diff
@@ -182,18 +187,25 @@ public boolean isLIstAdditionalClusterSuccess() {
...
+                            // replace the duplicated cluster with linked one
+                            .map(cluster -> hdinsightAdditionalClusterDetails.stream()
+                                                .filter(linkedCluster -> linkedCluster.getName().equals(cluster.getName()))
+                                                .findFirst()
+                                                .orElse(cluster))
+                            .collect(Collectors.toList());
+
+                    mergedClusters.addAll(emulatorClusterDetails);
 
+                    ClusterMetaDataService.getInstance().addCachedClusters(mergedClusters);
```
djlan: How about users sign in but link cluster not belonging to their subscription? It seems like we are not adding hdinsightAdditionalClusterDetails or emulatorClusterDetails if users signed in but do not have any subscription? Probably we should do a full-outer join here not left-outer join

wezhang: Good catch! Thanks DJ! And there is another issue found of not filtering the selected subscription.

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/serverexplore/ui/AddNewClusterFrom.java
```diff
@@ -85,75 +93,89 @@ public AddNewClusterFrom(final Project project, HDInsightRootModule hdInsightMod
...
+                errorMessage = "All (*) fields are required.";
+                isCarryOnNextStep = false;
+            } else {
+                clusterName = getClusterName(clusterNameOrUrl);
+
+                if (clusterName == null) {
+                    errorMessage = "Wrong cluster name or endpoint";
+                    isCarryOnNextStep = false;
+                } else {
+                    int status = ClusterManagerEx.getInstance().isHDInsightAdditionalStorageExist(clusterName, storageName);
```
djlan: >int status = ClusterManagerEx.getInstance().isHDInsightAdditionalStorageExist(clusterName, storageName); [](start = 20, length = 104)

As discussed, let's not require the cluster to be default cluster

Issue #1338, opened 2018-01-10
[Eclipse]Fail to submit jobs after reopening Eclipse without touching Azure Explorer on Mac and Linux


tankungx:
Steps:
1. Update the latest bits.
2. Not change workspace.
3. Submit job.

Actual:
Fail to submit job. work well after change the workspace.
List spark clusters ...
List spark clusters successfully
List additional spark clusters successfully
Info : Get target jar from /home/linux16/eclipsekunworkspace1/temp/temp.jar.
Info : Begin uploading file /home/linux16/eclipsekunworkspace1/temp/temp.jar to Azure Blob Storage Account wasbs://zhwe-spkdbg2@zhwespk.blob.core.windows.net/SparkSubmission/2018/00/10/08df7edc-b003-46b7-9b9d-c417a44cee4c/temp.jar ...
Info : Submit file to azure blob 'wasbs://zhwe-spkdbg2@zhwespk.blob.core.windows.net/SparkSubmission/2018/00/10/08df7edc-b003-46b7-9b9d-c417a44cee4c/temp.jar' successfully.
**Error : Failed to submit application to spark cluster. Exception : null**

voidfunction:
Repro this issue on my computer

OS:
Ubuntu 17.04

Eclipse Information:
![image](https://user-images.githubusercontent.com/6745693/34761845-7368838c-f620-11e7-8e5f-b330d420f314.png)

Exception:
![image](https://user-images.githubusercontent.com/6745693/34761823-57c03e36-f620-11e7-9fd7-84c9a8a50dd1.png)

wezhang:
reproduced, open the Eclipse with an opened workspace, submittion failed.

wezhang:
SparkBatchSubmission

```
    private SparkBatchSubmission() {
        String installID = HDInsightLoader.getHDInsightHelper().getInstallationId();
        String userAgentSource = SparkBatchSubmission.class.getClassLoader().getClass().getName().toLowerCase().contains("intellij")
                ? "Azure Toolkit for IntelliJ " : "Azure Toolkit for Eclipse ";
        userAgentName = userAgentSource + installID;
    }
```

`HDInsightLoader.getHDInsightHelper()` returns null.

wezhang:
After switching to `Azure Explorer`, and switching back, the issue can be mitigated.

PR #1361, opened 2018-01-30
Fix the HDInsightLoader uninitialized issue on Eclipse Mac/Linux

wezhang:
Sometime, the Azure Explorer is lazy initialized, which caused
HDInsightLoader not set.


wezhang:
To fix #1338

Issue #1539, opened 2018-03-08
[Intellij][hdinsight] get Spark job log Error: 404 Not Found


wezhang:
```
get Spark job log Error: 404 Not Found for https://zhwe-spk21.azurehdinsight.net/yarnui/null/stdout?start=0
com.gargoylesoftware.htmlunit.FailingHttpStatusCodeException: 404 Not Found for https://zhwe-spk21.azurehdinsight.net/yarnui/null/stdout?start=0
	at com.gargoylesoftware.htmlunit.WebClient.throwFailingHttpStatusCodeExceptionIfNecessary(WebClient.java:579)
```

wezhang:
Also sometimes got:

```
get Spark job log Error: 404 Not Found for https://zhwe-spk20.azurehdinsight.net/yarnui/wn1-zhwe-s.dlaibsjnhkiezoyqowrnnvwwld.dx.internal.cloudapp.net/node/containerlogs/container_1519369043841_0009_01_000001/livy/stdout?start=0
com.gargoylesoftware.htmlunit.FailingHttpStatusCodeException: 404 Not Found for https://zhwe-spk20.azurehdinsight.net/yarnui/wn1-zhwe-s.dlaibsjnhkiezoyqowrnnvwwld.dx.internal.cloudapp.net/node/containerlogs/container_1519369043841_0009_01_000001/livy/stdout?start=0
	at com.gargoylesoftware.htmlunit.WebClient.throwFailingHttpStatusCodeExceptionIfNecessary(WebClient.java:579)

```

PR #1540, opened 2018-03-08
Fix the get Spark job log not found error with null check

wezhang:


wezhang:
to fix #1539

Issue #1324, opened 2017-12-29
job failed with ClassNotFoundException when run configuration without mainclass


voidfunction:
1. Create a new project.
2. Create a run configuration without main class. And then save the configuration.
![image](https://user-images.githubusercontent.com/6745693/34435427-8c10bdde-ecc8-11e7-9d86-24dde0ba7a3c.png)
3. Click the run button without main class.
![image](https://user-images.githubusercontent.com/6745693/34435460-c2ee4330-ecc8-11e7-9ee1-5047ec6562a4.png)
4. Job failed with ClassNotFoundException.
```
LOG: AM container is launched, waiting for AM container to Register with RM
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/usr/hdp/2.5.5.3-2/spark2/jars/slf4j-log4j12-1.7.16.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/usr/hdp/2.5.5.3-2/hadoop/lib/slf4j-log4j12-1.7.10.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.Log4jLoggerFactory]
17/12/29 10:44:00 INFO SignalUtils: Registered signal handler for TERM
17/12/29 10:44:00 INFO SignalUtils: Registered signal handler for HUP
17/12/29 10:44:00 INFO SignalUtils: Registered signal handler for INT
17/12/29 10:44:01 INFO ApplicationMaster: Preparing Local resources
17/12/29 10:44:01 INFO MetricsConfig: loaded properties from hadoop-metrics2-azure-file-system.properties
17/12/29 10:44:01 INFO WasbAzureIaasSink: Init starting.
17/12/29 10:44:01 INFO AzureIaasSink: Init starting. Initializing MdsLogger.
17/12/29 10:44:01 INFO AzureIaasSink: Init completed.
17/12/29 10:44:01 INFO WasbAzureIaasSink: Init completed.
17/12/29 10:44:01 INFO MetricsSinkAdapter: Sink azurefs2 started
17/12/29 10:44:01 INFO MetricsSystemImpl: Scheduled snapshot period at 60 second(s).
17/12/29 10:44:01 INFO MetricsSystemImpl: azure-file-system metrics system started
[...]

PR #1547, opened 2018-03-27
Fix the Spark Job submission parameter checking issue

wezhang:


wezhang:
To fix #1324

Issue #1332, opened 2018-01-09
[Azure Toolkit for IntelliJ] Multi-Factor Authentication does not work


mirodd:
Along with my colleagues, we are having troubles when signing in from IntelliJ using Azure Explorer (part of Azure Toolkit for IntelliJ) to Azure.
We are using the latest version of the toolkit (3.1.0 – 2017.3) but we are getting a Multi-Factor Authentication error similar to this:

AD Auth token endpoint returned HTTP status code 400. Error info: {"error":"interaction_required","error_description":"AADSTS50076: Due to a configuration change made by your administrator, or because you moved to a new location, you must use multi-factor authentication to access 'xxx'.\r\nTrace ID: xxx\r\nCorrelation ID: xxx\r\nTimestamp: 2018-01-08 12:24:51Z","error_codes":[50076],"timestamp":"2018-01-08 12:24:51Z","trace_id":"xxx","correlation_id":"xxx"}

We have checked the issue tracker here GitHub but were not able to find anything related with that problem.
So I am opening a new issue and hope this to be solved as soon as possible.

Thanks.

Eskibear:
Current knowledge:

* The user can successfully sign-in with AAD (MFA passed)  in browser.

mirodd:
Hello,
We were having separate discussion on email but we are moving now here.
So, I've tried as suggested to log in using IE InPrivate tab or Chrome Incognito window via this link: https://login.microsoftonline.com/common/oauth2/authorize?return-client-request-id=true&resource=https%3A%2F%2Fmanagement.core.windows.net%2F&client-request-id=083f2439-b24a-4123-95da-af6b5f08c960&response_type=code&redirect_uri=https%3A%2F%2Fmsopentech.com%2F&prompt=login&client_id=61d65f5a-6e3b-468b-af73-a033f5098c5c

I am getting redirected properly to my company login page and when I put my credentials and then enter the code I get at my device, I am receiving an error I have previously sent (in the attached file)
![error_toolkit](https://user-images.githubusercontent.com/35266014/34814940-2d0b8d7e-f6b8-11e7-914d-6bb848bc645d.png)

On the other hand, I also contacted our IT administrators and no changes have been made to my account. In order to give you the full picture, I have managed to log on with the same account using an older version of the toolkit but do not remember the exact version. However, I need the new one as the previous was not working with HDInsight clusters when the default storage is Data Lake Store Account.

[...]

PR #1555, opened 2018-03-28
mitigation for issue #1332

Eskibear:
Unblock the users who cannot use `refresh_token` of `common` tenant to acquire other tenants' tokens.

Mitigation:
Pop up a dialog, allowing users to repeat the whole sign-in process on specific tenant.

### DIFF
--- a/Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/AdAuthManager.java
+++ b/Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/AdAuthManager.java
@@ -106,7 +106,7 @@ public class AdAuthManager {
 
         cleanCache();
         String commonTid = "common";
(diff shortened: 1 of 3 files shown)


Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/AdAuthManager.java
```diff
@@ -118,11 +118,16 @@ public AuthResult signIn() throws IOException {
             String tid = t.tenantId();
             AuthContext ac1 = createContext(tid, null);
             // put tokens into the cache
-            ac1.acquireToken(env.managementEndpoint(), false, userId, isDisplayable);
+            try {
```
Eskibear: critical code here. code formatting in other places.

Issue #1294, opened 2017-12-11
Revert fix for #1288 after intelliJ 2018 released


adashen:
https://github.com/Microsoft/azure-tools-for-java/pull/1293 fixed #1288. The root cause is an intelliJ bug. And the fix will be in 2018. (https://github.com/JetBrains/intellij-community/commit/df6a596e15e2ffb0c2e6b6b4be8c4af0ef096a00#diff-52c7fa7387b3775c006937597017b726) So add an issue to track the status.

vscjavaci:
Track internal issue by AB#1738847

PR #1556, opened 2018-03-28
Support IntelliJ 2018.1

jdneo:
Since IntelliJ released 2018.1, we need to update the build scripts.

Meanwhile, some of the codes also need to change due to the API change of the storage annotation:
- https://www.jetbrains.org/intellij/sdk/docs/basics/persisting_state_of_components.html#defining-the-storage-location
- https://github.com/JetBrains/intellij-community/blob/master/platform/projectModel-api/src/com/intellij/openapi/components/Storage.java

attribute `id` is removed

### DIFF
--- a/BuildDevint
+++ b/BuildDevint
@@ -42,7 +42,7 @@ chmod +x ./tools/IntellijVersionHelper
 if [ $INTELLIJ_VERSION == "true" ] ; then
     ./tools/IntellijVersionHelper 2017.3
 fi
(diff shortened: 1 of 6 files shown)


Eskibear:
you may also want to check this issue https://github.com/Microsoft/azure-tools-for-java/issues/1294

jdneo:
Emm, seems we still needs to leave the work around there since 2017.3 is still in support.

wezhang:
@jdneo , to avoid unsupported issue between new IDEA version and our release, maybe we can start the support and testing from the IDEA EAP phrase.

Issue #1549, opened 2018-03-27
[IntelliJ] continue debugging get errors


jingyanjingyan:
Repro Steps:
1. Set configure file and remote debug like "SparkCore_WebIOTest", success
2. Edit output, click Re-debug button, start to debugging

Result:
![image](https://user-images.githubusercontent.com/36756024/37956942-46582500-31df-11e8-8a11-fdbf45f23aa8.png)

wezhang:
Only occurred in password SSH auth type.

PR #1557, opened 2018-03-29
Enable Spark Job Debug SSH password saving with IDEA credential store

wezhang:


wezhang:
to fix #1549

Issue #1576, opened 2018-04-04
[IntelliJ] Open ASE for storage not works for linux and mac


jingyanjingyan:
Steps:
1. Right click on a cluster, click "Open Azure Storage Explorer for storage"

Result:
![image](https://user-images.githubusercontent.com/36756024/38285551-48b4e436-37f3-11e8-8fbe-d0211d21a547.png)

wezhang:
Fixed on Mac os.

v-jiche:
Verified. Fixed
artifacts/azure-toolkit-for-intellij-2018.1.PR.2083.09-19-2018.zip

PR #1579, opened 2018-04-04
Fix ASE deeplink with only opening the storage explorer

wezhang:


wezhang:
to fix #1576

PR #1584, opened 2018-04-08
using the latest AI dependency

jdneo:


Utils/pom.xml
```diff
@@ -342,12 +342,17 @@
...
+                <version>[2.0.0, 3.0.0)</version>
             </dependency>
             <dependency>
                 <groupId>com.microsoft.azure</groupId>
                 <artifactId>applicationinsights-web</artifactId>
-                <version>1.0.10</version>
+                <version>[2.0.0, 3.0.0)</version>
+            </dependency>
+            <dependency>
+                <groupId>commons-io</groupId>
```
Eskibear: already included in 3rd-party-notice?

jdneo: Yes we have: https://github.com/Microsoft/azure-tools-for-java/blob/develop/PluginsAndFeatures/AddLibrary/AzureLibraries/com.microsoft.azuretools.sdk/dependencies/ThirdPartyNotices.txt#L12

PR #1622, opened 2018-05-04
fix potential NPE: some locations have null value for region()

Eskibear:
e.g. new location like `france central`, `region()` returns `null`


jdneo:
@Eskibear how you found it? Just curious

wezhang:
We can add `@NotNull` and `@Nullable` as much as possible.

Issue #1577, opened 2018-04-04
Save the signed in session in Azure explorer


wezhang:
Need to leverage the credential manager to save tokens.

- [x] IntelliJ support
- [ ] Eclipse support

vscjavaci:
Track internal issue by AB#1738846

PR #1623, opened 2018-05-04
Introduce the interactive AD authentication persisting feature

wezhang:
Now, the interactive mode AD signed-in certificate can be saved in the IntelliJ secure store. The user needn't to re-sign in Azure when restarting IntelliJ.


wezhang:
to fix #1577 , IntelliJ part.

PR #1657, opened 2018-05-17
Add account Rest API support

t-rufang:
Error found when we use `@SkipParentValidation` in both `DataLakeAnalyticsAccount` class and `DataLakeAnalyticsAccountBasic` class. IntelliJ warns that `Cannot resolve symbol "SkipParentValidation"`.


wezhang:
Need azure-client v1.3.0, depends #1659

PR #1661, opened 2018-05-18
Enable provided environment settings.

wezhang:


Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/AdAuthManager.java
```diff
@@ -106,7 +106,9 @@ public String getAccessToken(String tid, String resource, PromptBehavior promptB
      * @return true for success
      */
     public synchronized boolean tryRestoreSignIn(@NotNull AuthMethodDetails authMethodDetails) {
-        if (secureStore == null) {
+        if (secureStore == null || authMethodDetails.getAzureEnv() == null ||
+                // Restore only for the same saved Azure environment with current
+                !CommonSettings.getEnvironment().getName().equals(authMethodDetails.getAzureEnv())) {
```
djlan: || (getAzureEnv != null && equals(getAzureEnv())?

Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/AdAuthManager.java
```diff
@@ -180,8 +182,22 @@ public AuthResult signIn(@Nullable AuthResult savedAuth) throws IOException {
...
                 ac1.acquireToken(env.managementEndpoint(), true, userId, isDisplayable);
             }
-            ac1.acquireToken(env.resourceManagerEndpoint(), false, userId, isDisplayable);
+
+            // FIXME!!! Some environments and subscriptions can't get the resource manager token
+            // Let the log in process passed, and throwing the errors when to access those resources
+            try {
+                ac1.acquireToken(env.resourceManagerEndpoint(), false, userId, isDisplayable);
+            } catch (AuthException ignored) {
+            }
```
djlan: CommonSettings.getEnvironment() instanceOf ProvidedEnvironment? Otherwise still throw exception?

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/HDIEnvironment.java
```diff
@@ -44,26 +44,25 @@ private HDIEnvironment(Map<String, String> endpoints, Environment environment) {
...
-                this.endpoints = HDIEnvironment.GERMANY.endpoints;
-                break;
-            case US_GOVERNMENT:
-                this.environment = Environment.US_GOVERNMENT;
-                this.endpoints = HDIEnvironment.US_GOVERNMENT.endpoints;
-                break;
-            default:
-                this.environment = Environment.GLOBAL;
-                this.endpoints = HDIEnvironment.GLOBAL.endpoints;
+        if (Environment.GLOBAL.equals(environment)) {
```
Eskibear: What's the reason to replace `switch case` with `if else`? To improve perf based on frequency of candidates?

wezhang: Since the `Environment` is changed from `Enum` to `Interface`, The values, such as `Environment.GLOBAL` can't be const to be used by `switch-case`, I have to replace them by `if-else`.

Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/AdAuthManager.java
wezhang: NP, fix it!

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/HDIEnvironment.java
wezhang: In fact, I prefer Polymorphism, but current change is minimal.

PR #1681, opened 2018-06-04
extend version range to include 182

Eskibear:


BuildDevint
```diff
@@ -54,4 +54,12 @@ fi
 
 cp ./PluginsAndFeatures/azure-toolkit-for-intellij/build/distributions/azure-toolkit-for-intellij.zip ./$ARTIFACTS_DIR/azure-toolkit-for-intellij-2018.1.zip
 
+# Build intellij 2018.1 plugin
```
jdneo: 2018.2?

PR #1690, opened 2018-06-11
Fix cluster state in run configuration panel not align with that in azure explorer issue

t-rufang:


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/spark/ui/ServerlessSparkSubmissionPanelConfigurable.kt
```diff
@@ -47,6 +47,7 @@ class ServerlessSparkSubmissionPanelConfigurable(private val project: Project, c
                 .doOnEach { submissionPanel.setClustersListRefreshEnabled(true) }
                 .subscribe(
                         { clusters ->
+                            clusters.forEach { cluster -> cluster.getConfigurationInfo() }
```
wezhang: It might be better to do refresh for each cluster in `.fetchClusters()` method with `.flatMap(AzureSparkServerlessCluster::get)`

PR #1692, opened 2018-06-12
use a private intellij-gradle-plugin build for SDK182

Eskibear:
Before JetBrains/gradle-intellij-plugin fixes SDK 182 issue, use a private one to build.


wezhang:
Is the private build just disable `instrumentCode` gradle task?

Eskibear:
@wezhang No, it's just the PR build that hasn't been merged by JetBrains. Simply added `forms-1.1-preview.jar` to classpath when execute `instrumentCode` task.

Skipping the task `instrumentCode` will cause error for all the UI composed by  `*.form`.

PR #1693, opened 2018-06-12
Fix the Livy Session API multiple UUID in UA issue

wezhang:


wezhang:
Checked the DB, only found `/livy/session` request.

PR #1728, opened 2018-06-19
 Fix clicking link in monitor dialog and poping up two instances in browser issue

t-rufang:


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/sparkserverless/common/JXHyperLinkWithUri.java
```diff
@@ -11,15 +11,6 @@
     @Nullable
     private URI uri;
 
-    public JXHyperLinkWithUri() {
-        super();
-        this.addActionListener(event -> {
-            if (this.uri != null) {
-                BrowserUtil.browse(this.uri);
```
wezhang: Removed?

t-rufang: In a word, we don't need to add open browser action for class `JXHyperlinkWithUri` since `setURI` method has alreay done that.
In `setURI` method of class `JXHyperLinkWithUri`, we can find `this.setAction(HyperlinkAction.createHyperlinkAction(uri))`. If we dive into the implementation of class `HyperlinkAction`, we will find in method `actionPerformed`, `this.getURIVisitor().visit((URI)this.getTarget());` will be executed.

Issue #1722, opened 2018-06-15
[Eclipse][Linux] the eclipse crashed when right click on menu list


liyanp1987:
**Repro bit**

WindowsAzurePlugin4EJ.develop.641.06-15-2018

**Repro Steps**

1. Sign in Eclipse and expand HDInsight
2. Choose any cluster
3. Right click and choose any one of menu list

**Actual result**

The Eclipse crashed

tankungx:
Can repro on another machine on Linux.

wezhang:
Could you attach your JDK version? looks like a JDK opening browser issue.

wezhang:
The JDK browse issue, crashed on here:

```
Desktop.getDesktop().browse(new URI(linkUrl));
```

wezhang:
The same issue of Oracle JDK:

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f147e7a42b7, pid=8899, tid=0x00007f14f5236700
#
# JRE version: Java(TM) SE Runtime Environment (8.0_171-b11) (build 1.8.0_171-b11)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.171-b11 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [libgdk-x11-2.0.so.0+0x4e2b7]  gdk_display_open+0x57
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /home/zhwe/tools/eclipse/hs_err_pid8899.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
```

wezhang:
Related issue: https://bugs.eclipse.org/bugs/show_bug.cgi?id=430736

liyanp1987:
verified on bits dev 649, this issue is fixed, thanks

t-rufang:
Close the bug since it's fixed.

PR #1730, opened 2018-06-19
Replace JDK browser with IDE integrated browser to open the link

wezhang:


wezhang:
There still lots of invoking ` Desktop.getDesktop().browse(new URI(linkUrl));`, which should be fixed later.

wezhang:
To fix #1722

PR #1756, opened 2018-07-04
Enable available AU in provision dialog

t-rufang:


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/rest/azure/serverless/spark/models/ApiVersion.java
```diff
@@ -24,4 +24,5 @@
 
 public class ApiVersion {
     public static final String VERSION = "2018-02-01-preview"; // Preview version
+    public static final String JOB_API_VERSION = "2016-11-01";
```
wezhang: There is a special ApiVersion in job api, otherwise, create one for it.

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/common/azure/serverless/AzureSparkServerlessAccount.java
```diff
@@ -101,6 +108,25 @@ public URI getUri() {
...
+    public Observable<Integer> getJobDegreeOfParallelism() {
+        return getJobList()
+                .flatMap(jobList -> Observable.from(jobList.value()))
+                .map(jobInfo -> jobInfo.degreeOfParallelism())
+                .defaultIfEmpty(0)
+                .reduce((a, b) -> a + b);
+    }
+
+    // TODO: handle job list pagination
+    public Observable<JobInfoListResult> getJobList() {
```
wezhang: The method name getJobs() is enough.

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/common/azure/serverless/AzureSparkServerlessAccount.java
```diff
@@ -101,6 +108,25 @@ public URI getUri() {
     // RestFUL API operations
     //
 
+    public Observable<Integer> getJobDegreeOfParallelism() {
+        return getJobList()
+                .flatMap(jobList -> Observable.from(jobList.value()))
```
wezhang: map() is enough?

Utils/hdinsight-node-common/src/com/microsoft/azure/sparkserverless/serverexplore/SparkServerlessClusterProvisionCtrlProvider.java
```diff
@@ -75,7 +75,10 @@ public static int getCalculatedAU(int masterCores,
     }
 
     public Observable<Integer> getAvailableAU() {
-        return Observable.just(0);
+        return account.getJobDegreeOfParallelism()
+                .subscribeOn(Schedulers.io())
+                .map(jobDegreeOfParallelism -> account.getMaxDegreeOfParallelism() < jobDegreeOfParallelism ? 0
+                        : account.getMaxDegreeOfParallelism() - jobDegreeOfParallelism);
     }
```
djlan: In the dialogue, call UpdateTotalAU first (although it might still return after updateAvailableAU) because availableAU depends on result of getTotalAU

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/common/azure/serverless/AzureSparkServerlessAccount.java
t-rufang: Firstly, the implementation of `values()`  method is as below.
```
    public List<JobInformationBasic> value() {
        return this.value;
    }
```

If we use `map` here, codes will look as below, which is strange to me to introduce java stream chain calls in rxjava chain calls.
```
        return getJobs()
                .map(jobList -> jobList.value().stream().mapToInt(jobInfo -> jobInfo.degreeOfParallelism()).sum())
                .defaultIfEmpty(0);
```
If we want to get rid of `stream`, we can replace `mapToInt()`  with `Lists.transform` , but I can't figure out a better way to sum up all integer values in java List without `stream`.

Therefore, I prefer the original `flatMap` way, which is much more clear and concise.

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/rest/azure/serverless/spark/models/ApiVersion.java
t-rufang: Double confirmed, there is no `ApiVersion` class in Job API though I find one in account API.

---
In reply to: [200133979](https://github.com/Microsoft/azure-tools-for-java/pull/1756#discussion_r200133979) [](ancestors = 200133979)

Utils/hdinsight-node-common/src/com/microsoft/azure/sparkserverless/serverexplore/SparkServerlessClusterProvisionCtrlProvider.java
```diff
@@ -75,7 +76,15 @@ public static int getCalculatedAU(int masterCores,
...
 
     public Observable<Integer> getAvailableAU() {
-        return Observable.just(0);
+        return account.getJobDegreeOfParallelism()
+                .subscribeOn(Schedulers.io())
+                .map(jobDegreeOfParallelism -> account.getMaxDegreeOfParallelism() < jobDegreeOfParallelism ? 0
+                        : account.getMaxDegreeOfParallelism() - jobDegreeOfParallelism);
+    }
+
+    public Observable<Pair<Integer, Integer>> getAvailableAUAndTotalAUFirstTime() {
```
djlan: >getAvailableAUAndTotalAUFirstTime [](start = 46, length = 33)

Rename to getAvailableAUAndTotalAU?

PR #1773, opened 2018-07-13
Fix the Azure Explorer Data Lake Spark node refresh issue

wezhang:


wezhang:
To fix #1771

Issue #1776, opened 2018-07-13
[IntelliJ][ReportedByUser] Uncaught Exception Cannot read scheme SubmitSparkJob_Configuration-_Spark Job_ SparkCore_WasbIOTest_ factoryName_ HDInsight Spark.xml: Illegal char <:> at index 13: C:\WinUtils;C:\Users\v-yajing\Desktop\HDI\hadoop-3.1.0\hadoop-3.1.0;\bin\winutils.exe java.nio.file.InvalidPathException: Illegal char <:> at index 13: C:\WinUtils;C:\Users\v-yajing\Desktop\HDI\hadoop-3.1.0\hadoop-3.1.0;\bin\winutils.exe


jingyanjingyan:
IntelliJ build version: 2018.1.3 IC-181.4892.42
JDK: JetBrains s.r.o 1.8.0_152-release
Plugin version: 3.8.0
Additional Info: None
Parent component: javax.swing.JPanel[,0,0,771x287,invalid,layout=java.awt.BorderLayout,alignmentX=0.0,alignmentY=0.0,border=javax.swing.border.EmptyBorder@1dd74546,flags=9,maximumSize=,minimumSize=,preferredSize=]
Error message: IdeaLoggingEvent[message=Cannot read scheme SubmitSparkJob_Configuration-_Spark Job_ SparkCore_WasbIOTest_ factoryName_ HDInsight Spark.xml: Illegal char <:> at index 13: C:\WinUtils;C:\Users\v-yajing\Desktop\HDI\hadoop-3.1.0\hadoop-3.1.0;\bin\winutils.exe, throwable=java.nio.file.InvalidPathException: Illegal char <:> at index 13: C:\WinUtils;C:\Users\v-yajing\Desktop\HDI\hadoop-3.1.0\hadoop-3.1.0;\bin\winutils.exe
	at sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)
	at sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:255)
	at java.nio.file.Paths.get(Paths.java:84)
[...]

PR #1798, opened 2018-07-16
Fix the error thrown when HADOOP_HOME env configurated wrong

wezhang:


wezhang:
To fix #1776

Issue #1774, opened 2018-07-13
[IntelliJ][ReportedByUser] MS Azure Explorer - Error Loading cosmos-partner-c14


jingyanjingyan:
Environment: Product
Repro Steps:
1. Refresh cluster "Cosmos-partner-c14"
Result:
Error dialog
![image](https://user-images.githubusercontent.com/36756024/42743815-33036100-88fa-11e8-82e3-c71ffd65d7a0.png)

	rx.exceptions.OnErrorNotImplementedException
		at rx.internal.util.InternalObservableUtils$ErrorNotImplementedAction.call(InternalObservableUtils.java:386)
		at rx.internal.util.InternalObservableUtils$ErrorNotImplementedAction.call(InternalObservableUtils.java:383)
		at rx.internal.util.ActionSubscriber.onError(ActionSubscriber.java:44)
		at rx.observers.SafeSubscriber._onError(SafeSubscriber.java:153)
		at rx.observers.SafeSubscriber.onError(SafeSubscriber.java:115)
		at rx.internal.operators.OnSubscribeFilter$FilterSubscriber.onError(OnSubscribeFilter.java:90)
		at rx.exceptions.Exceptions.throwOrReport(Exceptions.java:216)
		at rx.internal.operators.OperatorOnErrorResumeNextViaFunction$4.onError(OperatorOnErrorResumeNextViaFunction.java:144)
		at rx.internal.operators.OperatorSwitchIfEmpty$ParentSubscriber.onError(OperatorSwitchIfEmpty.java:84)
		at rx.internal.operators.OnSubscribeMap$MapSubscriber.onError(OnSubscribeMap.java:88)
		at rx.internal.operators.OnSubscribeMap$MapSubscriber.onError(OnSubscribeMap.java:88)
		at rx.internal.operators.OperatorMerge$MergeSubscriber.reportError(OperatorMerge.java:266)
		at rx.internal.operators.OperatorMerge$MergeSubscriber.checkTerminate(OperatorMerge.java:818)
		at rx.internal.operators.OperatorMerge$MergeSubscriber.emitLoop(OperatorMerge.java:579)
		at rx.internal.operators.OperatorMerge$MergeSubscriber.emit(OperatorMerge.java:568)
		at rx.internal.operators.OperatorMerge$InnerSubscriber.onError(OperatorMerge.java:852)
		at rx.observers.Subscribers$5.onError(Subscribers.java:230)
[...]

PR #1799, opened 2018-07-16
Fix the misuse of RxJava onErrorResumeNext

wezhang:
Should return an observable


wezhang:
to fix #1774

Issue #1810, opened 2018-07-17
[IntelliJ]Configurantions dialog shows slowly when add Azure HDInsight Spark


tankungx:
azure-toolkit-for-intellij-2018.1.PR.1757.07-13-2018.zip
Steps:
1.Create a new project
2.Click "Edit Configurantions..."
3.Click "+" then choose "Azure HDInsight Spark"

Actual result:
The corresponding dialog on the right pane show slowly(about 8 seconds)

jingyanjingyan:
Fix with market build

PR #1811, opened 2018-07-17
Fix the Run Configuration open slow issue

wezhang:
Only force to fresh the first time, use cased clusters then.


wezhang:
To fix #1810

Issue #1968, opened 2018-09-07
Update the table of listing web apps to better align runtime information between Windows web apps and Linux web apps


SummerSun:
Current the table shows information like this:

| Name |  JDK | Web Container   |  Resource Group | 
|--|---|---|---|

When we try to add the support for Linux web apps, we find that the JDK versions from SDK are inconsistent between Linux web apps and Windows web apps. And Java SE web app has no information of web container. It is better for us to update the table and show consistent runtime information to users.

| Name |  OS | Runtime   |  Resource Group | 
|--|---|---|---|

PR #1974, opened 2018-09-07
Combine JDK and Web Container into Runtime and show OS info of a web app

SummerSun:
Fix issue #1968 

Use table row `Runtime` to show a combined and refactored information of Linux or Windows web app.
Add new table row `OS` to show the operating system of a web app.

image

After some discussion about the format, updated the format of the Runtime:

image

### DIFF
--- a/Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
+++ b/Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
@@ -375,7 +375,8 @@ public class WebAppUtils {
      * Docker web apps are not included.
      */
     public static boolean isJavaWebApp(@NotNull WebApp webApp) {
-        return webApp.javaVersion() != JavaVersion.OFF || StringUtils.containsIgnoreCase(webApp.linuxFxVersion(), "jre8");
+        return (webApp.operatingSystem() == OperatingSystem.WINDOWS && webApp.javaVersion() != JavaVersion.OFF)
+         || (webApp.operatingSystem() == OperatingSystem.LINUX && StringUtils.containsIgnoreCase(webApp.linuxFxVersion(), "jre8"));
     }
 
     /**
@@ -385,17 +386,31 @@ public class WebAppUtils {
      * We return a combined and refactored information as Java Runtime.
      */
     public static String getJavaRuntime(@NotNull final WebApp webApp) {
-        final String delimiter = "  ";
+        String webContainer;
         switch (webApp.operatingSystem()) {
             case WINDOWS:
-                final String refactoredJavaVersion = "Java" + webApp.javaVersion().toString();
-                return String.join(delimiter, webApp.javaContainer(), webApp.javaContainerVersion(), refactoredJavaVersion);
+                webContainer = webApp.javaContainer() == null ? null : webApp.javaContainer().toLowerCase();
+                return String.format("%s %s (Java%s)",
+                    StringUtils.capitalize(webContainer), webApp.javaContainerVersion(), webApp.javaVersion().toString());
             case LINUX:
                 final String linuxVersion = webApp.linuxFxVersion();
                 if (linuxVersion == null) {
                     return DEFAULT_VALUE_WHEN_VERSION_INVALID;
+                }
+
+                // TOMCAT|8.5-jre8 -> Tomcat 8.5 (JRE8)
+                final String[] versions = linuxVersion.split("\\||-");
+                if (versions == null && versions.length != 3) {
+                    return linuxVersion;
+                }
+
+                webContainer = versions[0].toLowerCase();
+                final String webContainerVersion = versions[1];
+                final String jreVersion = versions[2];
+                if (webContainer.contains("tomcat")) {
+                    return String.format("%s %s (%s)", StringUtils.capitalize(webContainer), webContainerVersion, jreVersion.toUpperCase());
                 } else {
-                    return linuxVersion.replace("|", delimiter).replace("-", delimiter);
+                    return jreVersion.toUpperCase();
                 }
             default:
                 return DEFAULT_VALUE_WHEN_VERSION_INVALID;
--- a/Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
+++ b/Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
@@ -398,7 +398,6 @@ public class WebAppUtils {
                     return DEFAULT_VALUE_WHEN_VERSION_INVALID;
                 }
 
-                // TOMCAT|8.5-jre8 -> Tomcat 8.5 (JRE8)
                 final String[] versions = linuxVersion.split("\\||-");
                 if (versions == null && versions.length != 3) {
                     return linuxVersion;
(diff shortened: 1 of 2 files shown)


Eskibear:
For UI changes, suggest to put a screenshot of the new looking in description, that would be more clear I think.

Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
```diff
@@ -371,60 +371,34 @@ public ResourceGroup getRg() {
...
-     * And we will return N/A for those kind of web apps.
+     * For a Windows web app, APIs are separated to get jdk information and web container information.
+     * For a Linux web app, API app.linuxFxVersion() returns a combined information, like
+     * "Tomcat|8.5-jre8" if it is a Linux web app with the web container Tomcat.
+     * We return a combined and refactored information as Java Runtime.
      */
-    public static String getWebContainer(@NotNull final WebApp webApp) {
-        switch(webApp.operatingSystem()) {
+    public static String getJavaRuntime(@NotNull final WebApp webApp) {
+        final String delimiter = "  ";
```
Eskibear: Why two spaces here?

Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
```diff
@@ -371,60 +371,34 @@ public ResourceGroup getRg() {
...
-                if (versions.length != 2) {
-                    return linuxVersion;
-                }
-                if (StringUtils.containsIgnoreCase(versions[0], "tomcat")) {
-                    return versions[0].replace("|", " ");
+                if (linuxVersion == null) {
+                    return DEFAULT_VALUE_WHEN_VERSION_INVALID;
                 } else {
-                    return "N/A";
+                    return linuxVersion.replace("|", delimiter).replace("-", delimiter);
```
Eskibear: Why bother to format the value? Is it better to show what SDK returns?

Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
```diff
@@ -371,60 +371,34 @@ public ResourceGroup getRg() {
...
-                    return DEFAULT_VALUE_WHEN_VERSION_INVALID;
-                }
-
-                final String[] versions = linuxVersion.split("-");
-                return versions.length != 2 ? linuxVersion : versions[1];
-            default:
-                return DEFAULT_VALUE_WHEN_VERSION_INVALID;
-        }
+    public static boolean isJavaWebApp(@NotNull WebApp webApp) {
+        return webApp.javaVersion() != JavaVersion.OFF || StringUtils.containsIgnoreCase(webApp.linuxFxVersion(), "jre8");
```
Eskibear: I think it is updated based on the original conditional. Just one thought, is it precise and is there any better choice?

Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
SummerSun: Better readability.

Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
SummerSun: Are you suggesting that we use the format to combine the Windows Runtime? Like `Tomcat|8.5-Jre8`, I don't think it is a good format to show in our UI to users, especially the `|` between web container and its version. Maybe some suggestions about the formats?

Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
SummerSun: Really don't see any better choice from the limited APIs we could leverage. Maybe add the operating system check along with the version check to improve the readability, but not that necessary from the code logic actually.

Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
Eskibear: I prefer not to format the value returned by SDK. For me, it's natural to combine two values returned by SDK in some format, and I'd rather not to do some tricky things inside each of them. 
@jdneo Do you have some suggestion?

Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
Eskibear: > Maybe add the operating system check along with the version check to improve the readability,

Agree. Better readability leads to less pain in future maintaining work.

Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
Eskibear: Or can we leverage the format in Azure portal for consistence instead of creating a new format by our own? For Linux Web App, I see the format is "Tomcat 8.5 (JRE 8)".

Eskibear: And "JAVA 8 jre8" also looks wired. 🤦‍♂️

SummerSun: Agree to use the format `Tomcat 8.5 (JRE8)` for Linux web apps and `Tomcat 8.5 (Java1.8)` for Windows web apps.
For Java SE web app, we will just show `JRE8`.

Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
SummerSun: Cool

Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
```diff
@@ -385,17 +386,31 @@ public static boolean isJavaWebApp(@NotNull WebApp webApp) {
...
     public static String getJavaRuntime(@NotNull final WebApp webApp) {
-        final String delimiter = "  ";
+        String webContainer;
         switch (webApp.operatingSystem()) {
             case WINDOWS:
-                final String refactoredJavaVersion = "Java" + webApp.javaVersion().toString();
-                return String.join(delimiter, webApp.javaContainer(), webApp.javaContainerVersion(), refactoredJavaVersion);
+                webContainer = webApp.javaContainer() == null ? null : webApp.javaContainer().toLowerCase();
+                return String.format("%s %s (Java%s)",
+                    StringUtils.capitalize(webContainer), webApp.javaContainerVersion(), webApp.javaVersion().toString());
```
Eskibear: `StringUtils.capitalize(null)  = null`, is this what you expect if `webApp.javaContainer() == null` ?

Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
```diff
@@ -385,17 +386,31 @@ public static boolean isJavaWebApp(@NotNull WebApp webApp) {
...
+                }
+
+                webContainer = versions[0].toLowerCase();
+                final String webContainerVersion = versions[1];
+                final String jreVersion = versions[2];
+                if (webContainer.contains("tomcat")) {
+                    return String.format("%s %s (%s)", StringUtils.capitalize(webContainer), webContainerVersion, jreVersion.toUpperCase());
                 } else {
-                    return linuxVersion.replace("|", delimiter).replace("-", delimiter);
+                    return jreVersion.toUpperCase();
```
Eskibear: better add a comment as an example here.

Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
SummerSun: Yes, if any part of the format is invalid, we show it as it is, like what we do for Linux runtime.

PR #1981, opened 2018-09-11
Add API to get deployment slots of a web app

SummerSun:
API part preparation for supporting deployment slots.


Utils/azuretools-core/src/com/microsoft/azuretools/core/mvp/model/webapp/AzureWebAppMvpModel.java
```diff
@@ -53,6 +38,20 @@
 import com.microsoft.azuretools.core.mvp.model.AzureMvpModel;
 import com.microsoft.azuretools.core.mvp.model.ResourceEx;
 import com.microsoft.azuretools.utils.WebAppUtils;
+import java.io.File;
```
Eskibear: Don't change import order. Defined as:
`STATIC###STANDARD_JAVA_PACKAGE###SPECIAL_IMPORTS###THIRD_PARTY_PACKAGE`

PR #1988, opened 2018-09-12
Add slot icons

SummerSun:


Eskibear:
please reference the PR where related code belongs.

Eskibear:
...otherwise it might take effort to find out why the icons are added in future.

PR #1996, opened 2018-09-13
List all app service plan in select subscription

SummerSun:
Since there is no the limitation that the web app and its service plan should belong to the same resource group. We should list all app service plan of the selected subscriptions for users to select.


SummerSun:
Fix issue #1993

PR #1997, opened 2018-09-13
Correct conditions to toggle DeployToRoot checkbox

SummerSun:
Only when web container is enabled in the web app, the checkbox `Deploy to Root` should be enabled.

pr

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/webapp/webappconfig/ui/WebAppSettingPanel.java
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/webapp/webappconfig/ui/WebAppSettingPanel.java
@@ -47,6 +47,7 @@ import com.microsoft.azuretools.core.mvp.model.ResourceEx;
 import com.microsoft.azuretools.core.mvp.model.webapp.JdkModel;
 import com.microsoft.azuretools.utils.WebAppUtils;
 import com.microsoft.intellij.runner.AzureSettingPanel;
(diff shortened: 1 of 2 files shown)


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/webapp/webappconfig/ui/WebAppSettingPanel.java
```diff
@@ -539,18 +539,28 @@ private void toggleRuntimePanel(final boolean isWindows) {
...
-        lblWebContainer.setVisible(isWar && rdoWindowsOS.isSelected());
-        cbWebContainer.setVisible(isWar && rdoWindowsOS.isSelected());
-        chkToRoot.setVisible(isWar && isWebContainerAvailable());
-        lblJarDeployHint.setVisible(!isWar && rdoUseExist.isSelected());
+    private void toggleJarDeployHint(final boolean isDeployingWar) {
+        lblJarDeployHint.setVisible(!isDeployingWar && rdoUseExist.isSelected());
     }
 
-    private boolean isWebContainerAvailable() {
+    private boolean isAbleToDeployToRoot() {
```
jdneo: Why checking the OS to determine if is able to deploy to root instead of checking the deployment file.

SummerSun: For Windows Web App, web container is always enabled. But for Linux Web App, Java SE web app has no web container defined.

PR #2017, opened 2018-09-17
Do not freeze UI when getting deployment slots

SummerSun:


SummerSun:
All `RefreshableNode` load function use background thread to refresh items.
Abadon this PR  to keep the behavior align with other modules/nodes.

PR #2018, opened 2018-09-17
Rename jar artifact to app.jar for Linux Java SE web app

SummerSun:
Linux Java SE web app requires the artifact to be named as `app.jar` to run.

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/webapp/webappconfig/WebAppRunState.java
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/webapp/webappconfig/WebAppRunState.java
@@ -36,6 +36,7 @@ import com.microsoft.azuretools.utils.AzureUIRefreshEvent;
 import com.microsoft.azuretools.utils.WebAppUtils;
 import com.microsoft.intellij.runner.AzureRunProfileState;
 import com.microsoft.intellij.runner.RunProcessHandler;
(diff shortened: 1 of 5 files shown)


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/webapp/webappconfig/WebAppRunState.java
```diff
@@ -36,6 +36,7 @@
 import com.microsoft.azuretools.utils.WebAppUtils;
 import com.microsoft.intellij.runner.AzureRunProfileState;
 import com.microsoft.intellij.runner.RunProcessHandler;
+import com.microsoft.intellij.runner.container.utils.Constant;
```
Eskibear: Wrong place for the constant `LINUX_JAVA_SE_RUNTIME`. Should not be placed under `com.microsoft.intellij.runner.container`.

Utils/azuretools-core/src/com/microsoft/azuretools/Constants.java
```diff
@@ -32,4 +32,5 @@
     public static String clientId = "61d65f5a-6e3b-468b-af73-a033f5098c5c";
```
Eskibear: I don't think here is the best place for the string, global constants related with sign-in are located here. For this webapp related constant, let's put it into `/com/microsoft/intellij/runner/webapp/webappconfig` or `/com/microsoft/intellij/runner/webapp`

BTW, Remove the comments above this class for legal concern.

PR #2028, opened 2018-09-18
Revert WebAppModulePresenter onModuleRefresh function

SummerSun:
Keep the function run in Module load thread in order to align the refresh behavior with other modules or nodes.

### DIFF
--- a/Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/webapp/WebAppModulePresenter.java
+++ b/Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/webapp/WebAppModulePresenter.java
@@ -1,17 +1,9 @@
 package com.microsoft.tooling.msservices.serviceexplorer.azure.webapp;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
 
-import com.microsoft.azure.management.appservice.WebApp;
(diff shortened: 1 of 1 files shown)


Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/webapp/WebAppModulePresenter.java
```diff
@@ -4,36 +4,19 @@
...
     public void onModuleRefresh() {
-        Observable.fromCallable(() -> AzureWebAppMvpModel.getInstance().listAllWebApps(true))
-            .subscribeOn(getSchedulerProvider().io())
-            .subscribe(webApps -> DefaultLoader.getIdeHelper().invokeLater(() -> {
-                if (isViewDetached()) {
-                    return;
-                }
-                getMvpView().renderChildren(webApps);
-            }), e -> errorHandler(CANNOT_GET_WEB_APPS, (Exception) e));
+        final WebAppModuleView view = getMvpView();
```
Eskibear: The whole block can be written as:
```java
if (!isViewDetached()) {
    getMvpView().renderChildren(AzureWebAppMvpModel.getInstance().listAllWebApps(true));
}
```

This looks simpler to me, but I'm not sticking to it.

SummerSun: Will resolve after fixing the block issues.

PR #2032, opened 2018-09-18
Show Linux web apps or slots with different icons

SummerSun:
Update the icon path to the patter `capitalizeOperatingSystem` + `iconRunningPostfix` and use different icons for Linux web app/slots and Windows web apps/slots.


Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/webapp/WebAppNode.java
```diff
@@ -203,11 +209,11 @@ public void renderWebAppNode(@NotNull WebAppState state) {
         switch (state) {
             case RUNNING:
                 this.webAppState = state;
-                this.setIconPath(this.ICON_RUNNING);
+                this.setIconPath(this.webAppOS + ICON_RUNNING_POSTFIX);
```
Eskibear: Is it possible to add a helper to generate icon path from webapp? e.g. 
```java
private String iconPath(WebApp app) {
    app.os .....
    app.state ....
    return  ....
}
```

Eskibear: Or if we don't want to hold the  web app as a member, it should be better to use a unified helper instead of concatenating the path everywhere.

SummerSun:
Track the first item in issue #2040

PR #1889, opened 2018-08-22
fix NoSuchFieldError

jdneo:
fix #1885

Keep align with the azure sdk: https://github.com/Azure/azure-libraries-for-java/blob/master/pom.xml#L59-L70

PR #2099, opened 2018-09-26
Upgrade Azure SDK to 1.16.0

SummerSun:
Upgrade to the latest Azure SDK 1.16.0. Reference this [pom.xml](https://github.com/Azure/azure-libraries-for-java/blob/v1.16.0/pom.xml)

#2103


Eskibear:
@jdneo Do we need to update other files? Refer to: #1893

jdneo:
Yes, some other files are also needed to change. My fault.

jdneo:
There are several other places that may need to update
- https://github.com/Microsoft/azure-tools-for-java/blob/develop/PluginsAndFeatures/azure-toolkit-for-intellij/build.gradle#L80-L81
- https://github.com/Microsoft/azure-tools-for-java/pull/1889/files
- https://github.com/Microsoft/azure-tools-for-java/pull/1891/files

SummerSun:
https://github.com/Microsoft/azure-tools-for-java/pull/1889/files is to update the `jackson-annotations`, `jackson-core` etc. dependencies versions, do we need to update it too?

SummerSun:
If we reference this [pom.xml](https://github.com/Azure/azure-libraries-for-java/blob/v1.16.0/pom.xml) and the changes made in this file https://github.com/Microsoft/azure-tools-for-java/pull/1889/files, these three depencies versions are remain unchanged:
`jackson-annotations`  2.9.0
`jackson-core`  2.9.4
`jackson-databind`  2.9.4

But the following two are no longer in the dependency list:
`jackson-dataformat-xml` 2.9.4
`jackson-module-scala_2.11` 2.9.4

wezhang:
Separate HDInsight pom exclusions issue into #2107

Issue #2046, opened 2018-09-19
Add support to operate on deployment slots


SummerSun:
Operations on deployment slots include:
- [ ] Start a deployment slot
- [ ] Stop a deployment slot
- [ ] Restart a deployment slot
- [ ] Swap with production
- [ ] Open a deployment slot URL in a browser
 - [ ] Refresh the deployment slot node to sync its state running/stop

SummerSun:
Hide the deployment slots node until all operations are ready.

Eskibear:
Remember to revert #2062 after all the related features are implemented.

Eskibear:
You may want to update the description (the checkbox items)?

PR #2114, opened 2018-09-28
Add deployment slot operations

SummerSun:
#2046 
Add operations for deployment slots:
- Start
- Stop
- Restart
- Swap with production
- Refresh
  - Refresh the real state(running/stop) of the deployment slot


Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/webapp/deploymentslot/DeploymentSlotNode.java
```diff
@@ -32,24 +39,87 @@
...
+                DefaultLoader.getUIHelper().openInBrowser("http://" + hostName);
+            }
+        });
+
+        super.loadActions();
+    }
+
+    @Override
+    protected void onNodeClick(NodeActionEvent e) {
+        // RefreshableNode refresh itself whenever the node is clicked.
```
Eskibear: I remember it only refreshes on the first time being clicked?

Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/webapp/deploymentslot/DeploymentSlotNode.java
```diff
@@ -32,24 +39,87 @@
...
+        // The deployment slot node is just a single node for the time being.
+        // Override the function to do noting to disable the auto refresh functionality.
+    }
+
+    private void start() {
+        try {
+            presenter.onStartDeploymentSlot(this.subscriptionId, this.webAppId, this.slotName);
+        } catch (IOException e) {
+            e.printStackTrace();
```
Eskibear: What about the error handling? I don't thin doing nothing is a correct way, because if exception occurs, the app is not started. The user should be at least notified somehow.

Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/webapp/deploymentslot/DeploymentSlotNodePresenter.java
```diff
@@ -24,26 +24,52 @@
...
+        if (!isViewDetached()) {
+            view.renderNode(WebAppBaseState.fromString(slot.state()));
+        }
     }
 
-    public void onSwapWithProduction() {
-        // todo
+    public void onSwapWithProduction(final String subscriptionId, final String webAppId,
+                                     final String slotName) throws IOException {
+        AzureWebAppMvpModel.getInstance().swapSlotWithProduction(subscriptionId, webAppId, slotName);
```
Eskibear: How does the user know whether the operation is successful or not ?

SummerSun: The notification "Swapping with production" keeps showing in the notification bar until the operation finished. Just like the Restart operation.

Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/webapp/deploymentslot/DeploymentSlotNode.java
SummerSun: I will add a todo item here and let's discuss the error handling for both the deployment node operations and web app node operations.

Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/webapp/deploymentslot/DeploymentSlotNode.java
SummerSun: Yes, verified it again. My fault. But isn't it a little weird for a single node refreshes its state on the first time being clicked? The state is supposed to be fetched just when the node is created. The behavior looks like this:

comment

Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/webapp/deploymentslot/DeploymentSlotNodePresenter.java
Eskibear: " open the URL"  is aggressive to me. I think the notification is enough.

SummerSun:
Even though we have the `// todo`` items in our code, it is better for us to creat an issue to track this. https://github.com/Microsoft/azure-tools-for-java/issues/2126

Issue #2082, opened 2018-09-25
Support to deploy to deployment slots


SummerSun:
Together with listing the deployment slot in Azure Explorer, we will support to deploy to deployment slots.

When a web app is selected, all its deployment slots plus the **production** will be listed in a dropdown list to select to deploy.  The **production** means the web app itself, will show as the first element. Add if there are no deployment slots, the **production** will be the only available element.

![image](https://user-images.githubusercontent.com/5179063/45992890-a1763d00-c0be-11e8-9667-13f3d465dde4.png)

**production** is a reserved slot name, users could not create a deployment slot named production.

PR #2115, opened 2018-09-28
Update UI to support to deploy to slot

SummerSun:
#2082 
UI part of supporting to deploy to slot:

pr

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/webapp/webappconfig/ui/WebAppDeployMvpView.java
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/webapp/webappconfig/ui/WebAppDeployMvpView.java
@@ -40,7 +40,9 @@ import java.util.List;
 
 public interface WebAppDeployMvpView extends MvpView {
 
-    void renderWebAppsAndSlots(@NotNull List<ResourceEx<WebApp>> webAppLists);
+    void renderWebAppsTable(@NotNull List<ResourceEx<WebApp>> webAppLists);
+
+    void enableDeploymentSlotPanel();
 
     void fillSubscription(@NotNull List<Subscription> subscriptions);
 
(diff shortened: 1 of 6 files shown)


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/webapp/webappconfig/ui/WebAppDeployMvpView.java
```diff
@@ -40,7 +40,7 @@
 
 public interface WebAppDeployMvpView extends MvpView {
 
-    void renderWebAppsTable(@NotNull List<ResourceEx<WebApp>> webAppLists);
+    void renderWebAppsAndSlots(@NotNull List<ResourceEx<WebApp>> webAppLists);
```
Eskibear: Why do we have to change the interface?

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/webapp/webappconfig/ui/WebAppSettingPanel.java
```diff
@@ -496,11 +508,18 @@ public void renderWebAppsTable(@NotNull List<ResourceEx<WebApp>> webAppLists) {
                     table.setRowSelectionInterval(i, i);
                 }
             }
+            // render the deployment slot panel
```
Eskibear: I think deployment slots can be rendered on selecting a web app.

SummerSun: The deployment slot list is rendered on selecting a web app. But before any web app is loaded, we may disable the `Deploy to Slot` checkbox and the list. Cause it makes no sense for users to play with two items without any  web apps ready. What do you think?

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/webapp/webappconfig/ui/WebAppDeployMvpView.java
SummerSun: The logic here is that: the deploy slot panel depends on the web apps table is ready. Enabling the deploy slot panel without any web apps ready, or even no web apps in the selected subscription is weird for users. 

So when the presenter finishes loading web apps, it will render the web app table and the slot panel in this one function call. `renderWebAppsAndSlots` describes more clear what the function in charge of.

Issue #2122, opened 2018-09-29
Error occurred while stopping the running process of application insight project


bsaby:
Build: Daily534
OS: Windows/Mac/Linux

Repro steps:
1. Create a web project
2. Right-click project and configure the AI
[how to configure application insight in IntelliJ]( https://microsoft.sharepoint.com/teams/WPTFundamental/_layouts/OneNote.aspx?id=%2Fteams%2FWPTFundamental%2FShared%20Documents%2FAAPTServicing%2FAAPT%20Webtools%20Offshore%20Current%20Release&wd=target%28Java%20Tools%2FAzure%20Java%20Toolkits%2FAzure%20java%20toolkits%28IDE%5C%29-manual.one%7C7FAF569A-EA05-4B92-957A-626D02B04207%2F%5BIntelliJ%5DConfigure%20Application%20Insights%7C7A05FB47-67D3-4421-ACD0-46833E70E0FB%2F%29onenote:https://microsoft.sharepoint.com/teams/WPTFundamental/Shared%20Documents/AAPTServicing/AAPT%20Webtools%20Offshore%20Current%20Release/Java%20Tools/Azure%20Java%20Toolkits/Azure%20java%20toolkits(IDE)-manual.one#%5bIntelliJ%5dConfigure%20Application%20Insights&section-id={7FAF569A-EA05-4B92-957A-626D02B04207}&page-id={7A05FB47-67D3-4421-ACD0-46833E70E0FB}&end   )
3. Run the application
4. Stop the running process after the artifact is deployed successfully

**Result:**
Server log:
```
D:\Software\apache-tomcat-8.5.32\bin\catalina.bat run
[2018-09-29 03:10:10,748] Artifact helloworld2: Waiting for server connection to start artifact deployment...
Using CATALINA_BASE:   "C:\Users\v-sunxi\.IntelliJIdea2018.2\system\tomcat\Unnamed_helloworld2"
Using CATALINA_HOME:   "D:\Software\apache-tomcat-8.5.32"
Using CATALINA_TMPDIR: "D:\Software\apache-tomcat-8.5.32\temp"
Using JRE_HOME:        "C:\Program Files\Java\jdk1.8.0_181"
Using CLASSPATH:       "D:\Software\apache-tomcat-8.5.32\bin\bootstrap.jar;D:\Software\apache-tomcat-8.5.32\bin\tomcat-juli.jar"
[...]

PR #2134, opened 2018-09-30
Keep application insights dependencies version to fix regression

SummerSun:
After upgrade Azure SDK to v1.16.0, the `com.microsoft.azure:applicationinsights-web` are upgraded to v2.2.0 and it causes regression #2122.

Issue https://github.com/Microsoft/ApplicationInsights-Java/issues/755 was created in their side to track the issue. Now we keep using the 2.1.2 as our workaround.


SummerSun:
Close the PR to use the workaround application insights team proposed to fix the regression.

Issue #2142, opened 2018-10-05
[IntelliJ][ReportedByUser] The report feature itself is too pushy


haokanga:
IntelliJ build version: 2018.1.3 IU-181.4892.42
OS: Mac OS X
JDK: JetBrains s.r.o 1.8.0_152-release
Plugin version: 3.10.0-2018.1

#### 1. Are you satisfied with the Azure Spark development functionalities & experiences offered in Azure Toolkit for IntelliJ? Please rate us from 1 (poor) to 10 (excellent).
1

#### 2. What did you like or dislike about Azure Toolkit for IntelliJ?
The IntelliJ Event log is supposed to track critical events, and it should complain aloud only when things go wrong. There are several times that the event log window drew my attention but it turns out to be this survey request.

```
11:02 AM	We're listening
				Thanks for helping Microsoft improve Azure Toolkit experience!
				Your feedback is important. Please take a minute to fill out our
				user satisfaction survey
```

To be honest, I understand that you want to user feedback but asking the users to fill the survey each time they open IntelliJ is way too pushy. 

#### 3. What enhancements would you like Azure Toolkit for IntelliJ to implement?
Limit the frequency of this survey request, and give the user the option to disable it.

Eskibear:
Thanks for the feedback, we will improve the approach in the coming version.

For the moment, [here ](https://github.com/Microsoft/azure-tools-for-java/issues/1840#issuecomment-410948559)is a workaround to disable it. 

Also cc @wezhang for awareness.

wezhang:
Thanks @haokanga ! It's fixed.

Issue #1871, opened 2018-08-08
[IntelliJ][ReportedByUser] User feedback


santharamselva:
Irritating popup every time open the IDE, even though given comments. How many comments I need to give to stop getting the popup?

IntelliJ build version: 2018.2.1 IU-182.3911.36
JDK: JetBrains s.r.o 1.8.0_152-release
Plugin version: 3.9.0-2018.2

# 1. Are you satisfied with the Azure Spark development functionalities & experiences offered in Azure Toolkit for IntelliJ? Please rate us from 1 (poor) to 10 (excellent).

# 2. What did you like or dislike about Azure Toolkit for IntelliJ?

# 3. What enhancements would you like Azure Toolkit for IntelliJ to implement?

wezhang:
Thanks! You can refer to #1840 as a workaround to disable the popup before we provide the specific setting.

Issue #1840, opened 2018-07-27
[IntelliJ][ReportedByUser][Workaround] User feedback - annoyed feedback popup


hangryigor:
IntelliJ build version: 2018.1.6 IU-181.5540.7
JDK: JetBrains s.r.o 1.8.0_152-release
Plugin version: 3.9.0-2018.1

One thing to mention only: I did not even started with this plugin but I am already annoyed with this satisfaction survey message that pops up every time I start IntelliJ.

Eskibear:
Thanks for your feedback. 
@wezhang Could you help to take a look? Does the survey pops up with a proper sampling rate or not?

wezhang:
@hangryigor We'll add a setting page to configure it.

Before that, here is a workaround, you can follow these steps to disable it:

1. Open `Appearance & Behavior > Notifications` setting, Refer to https://www.jetbrains.com/help/idea/settings-notifications.html

2. Select `Azure Toolkit plugin` group, and change the Popup level to `No popup`. If you don't like the event is logged in Event Log window, deselect `Log` check box  please.
![image](https://user-images.githubusercontent.com/7385667/43758355-6e2adb5e-9a4e-11e8-89d5-6723f9de5cf0.png)

wezhang:
Close it since no new comments.

PR #2148, opened 2018-10-09
Improve user experience of the survey notification. 

Eskibear:
Previously the survey notification occurs whenever a project is open, which is complained by users in #1840 #1871 #2142 

This PR includes these changes:
* Change the survey notification to a one-time thing, it only pops up once the plugin is upgraded to a new version. 
* Add an entry in Tools menu for the survey, users can are able to provide feedback if they want to. 

image

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azuretools/ijidea/actions/GithubSurveyAction.java
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azuretools/ijidea/actions/GithubSurveyAction.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) Microsoft Corporation
+ *
+ * All rights reserved.
+ *
+ * MIT License
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+ * documentation files (the "Software"), to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and
+ * to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all copies or substantial portions of
+ * the Software.
+ *
+ * THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO
+ * THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package com.microsoft.azuretools.ijidea.actions;
+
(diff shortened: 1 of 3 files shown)


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/AzurePlugin.java
```diff
@@ -96,6 +94,8 @@
 
     private String _hashmac = GetHashMac.GetHashMac();
 
+    private Boolean firstInstallationByVersion;
```
wezhang: Unboxed `boolean` should be enough.

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azuretools/ijidea/actions/GithubSurveyAction.java
```diff
@@ -0,0 +1,40 @@
...
+
+package com.microsoft.azuretools.ijidea.actions;
+
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.microsoft.azuretools.ijidea.utility.AzureAnAction;
+import com.microsoft.intellij.feedback.GithubIssue;
+import com.microsoft.intellij.feedback.ReportableSurvey;
+
+
+public class GithubSurveyAction extends AzureAnAction {
```
wezhang: The class can inherit `NewGithubIssueAction` with `new GithubIssue(...)` in constructor.

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/AzurePlugin.java
wezhang: You need a lazy var? Looks like calling `isFirstInstallationByVersion()` is enough.

Eskibear: In previous code, `isFirstInstallationByVersion()` always tries to read "data.xml". But "data.xml" will be generated/updated by `initializeTelemetry()`. After this statement, `isFirstInstallationByVersion()` always returns `false`. That's why I add a field member here, to calculate and preserve the real value at the beginning of `initComponent()`.

Eskibear: As for the boolean vs Boolean, I need 3 states (null/true/false) to distinguish ( not yet compare data.xml / first-time installation / not first-time ), that's my intention to use the `Boolean`.
I think it's also ok to use the primitive type `boolean`, if we can make sure to assign the correct value to it at the very beginning.

wezhang: Cool, I image it might be better to add keyword `synchronized` to method `isFirstInstallationByVersion()` to avoid double-entry.

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azuretools/ijidea/actions/GithubSurveyAction.java
wezhang: Yes, by default is final in Kotlin. 😛  Just feel free to add the `open` keyword before class `NewGithubIssueAction` declaration.

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/AzurePlugin.java
Eskibear: added

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azuretools/ijidea/actions/GithubSurveyAction.java
Eskibear: updated.

PR #2149, opened 2018-10-09
Add place holder to remind users when no slot to deploy

SummerSun:
Remind the users that when there is no slot to deploy and we are going to deploy the web app instead.

comment


jdneo:
Why we do not have a default 'slot' which represents the Web App itself

SummerSun:
Use a placeholder `<No deployment slot is found, will deploy the web app.>` to remind users that there is no deployment slot and we are going to deploy the web app instead.

Eskibear:
Recap:

> Why we do not have a default 'slot' which represents the Web App itself

From the GIF, I see `staging` and `test` slots in a web app. So how do I deploy to the web app itself?

Eskibear:
> Recap:
> 
> > Why we do not have a default 'slot' which represents the Web App itself
> 
> From the GIF, I see `staging` and `test` slots in a web app. So how do I deploy to the web app itself?

Oh I see, am I supposed to uncheck the checkbox "Deploy to slot"?

SummerSun:
@Eskibear  Yep

SummerSun:
This PR discussed two options for reminding users that we are going to deploy the web app when there are no deployment slots:
- Show the information message at the bottom of the dialog
- Use the placeholder in the deployment slot combo box

But the above two options are kind of workarounds. Providing the option to create a deployment slot to deploy when there is none makes the scenario a completed one. So we close this PR and finish the scenario in another one to support the option for creating if not existing.

Issue #2172, opened 2018-10-10
[Intellij] Memory leak when shutting down Intellij after adding actions from SparkJobRun and SparkJobDebug


wezhang:
```
2018-10-10 00:50:34,476 [2143699]  ERROR - api.util.objectTree.ObjectTree - Memory leak detected: 'com.intellij.execution.impl.ConsoleViewImpl[,389,1,678x372,invalid,layout=java.awt.BorderLayout,alignmentX=0.0,alignmentY=0.0,border=,flags=9,maximumSize=,minimumSize=,preferredSize=]' of class com.intellij.execution.impl.ConsoleViewImpl
See the cause for the corresponding Disposer.register() stacktrace:
 
java.lang.RuntimeException: Memory leak detected: 'com.intellij.execution.impl.ConsoleViewImpl[,389,1,678x372,invalid,layout=java.awt.BorderLayout,alignmentX=0.0,alignmentY=0.0,border=,flags=9,maximumSize=,minimumSize=,preferredSize=]' of class com.intellij.execution.impl.ConsoleViewImpl
See the cause for the corresponding Disposer.register() stacktrace:

	at com.intellij.openapi.util.objectTree.ObjectTree.assertIsEmpty(ObjectTree.java:247)
	at com.intellij.openapi.util.Disposer.assertIsEmpty(Disposer.java:146)
	at com.intellij.openapi.util.Disposer.assertIsEmpty(Disposer.java:142)
	at com.intellij.openapi.application.impl.ApplicationImpl.disposeSelf(ApplicationImpl.java:251)
	at com.intellij.openapi.application.impl.ApplicationImpl.doExit(ApplicationImpl.java:790)
	at com.intellij.openapi.application.impl.ApplicationImpl.exit(ApplicationImpl.java:766)
	at com.intellij.openapi.application.impl.ApplicationImpl.exit(ApplicationImpl.java:755)
	at com.intellij.openapi.application.impl.ApplicationImpl.exit(ApplicationImpl.java:751)
	at com.intellij.openapi.application.impl.ApplicationImpl.exit(ApplicationImpl.java:718)
	at com.intellij.openapi.application.impl.ApplicationImpl.exit(ApplicationImpl.java:713)
	at com.intellij.openapi.wm.impl.IdeFrameImpl$3.windowClosing(IdeFrameImpl.java:275)
	at java.awt.AWTEventMulticaster.windowClosing(AWTEventMulticaster.java:349)
[...]

Issue #2167, opened 2018-10-10
[Intellij] Argument for @NotNull parameter 'action' of com/intellij/openapi/actionSystem/impl/ActionManagerImpl.getId must not be null


wezhang:
```
2018-10-10 11:10:39,693 [ 207638]  ERROR - api.util.objectTree.ObjectNode - Argument for @NotNull parameter 'action' of com/intellij/openapi/actionSystem/impl/ActionManagerImpl.getId must not be null 
java.lang.IllegalArgumentException: Argument for @NotNull parameter 'action' of com/intellij/openapi/actionSystem/impl/ActionManagerImpl.getId must not be null
	at com.intellij.openapi.actionSystem.impl.ActionManagerImpl.$$$reportNull$$$0(ActionManagerImpl.java)
	at com.intellij.openapi.actionSystem.impl.ActionManagerImpl.getId(ActionManagerImpl.java)
	at com.intellij.openapi.actionSystem.DefaultActionGroup.remove(DefaultActionGroup.java:226)
	at com.intellij.execution.ExecutorRegistryImpl.unregisterAction(ExecutorRegistryImpl.java:95)
	at com.intellij.execution.ExecutorRegistryImpl.deinitExecutor(ExecutorRegistryImpl.java:89)
	at com.intellij.execution.ExecutorRegistryImpl.dispose(ExecutorRegistryImpl.java:177)
	at com.intellij.openapi.util.Disposer$1.execute(Disposer.java:48)
	at com.intellij.openapi.util.Disposer$1.execute(Disposer.java:44)
	at com.intellij.openapi.util.objectTree.ObjectNode$1.execute(ObjectNode.java:138)
	at com.intellij.openapi.util.objectTree.ObjectNode$1.execute(ObjectNode.java:107)
```

wezhang:
There is ExecutorAction missing for SparkJobRun executors.

jingyanjingyan:
Verify with build develop 788 as fixed. Close IntelliJ and no errors in log.

PR #2175, opened 2018-10-10
Fix action null, memory leak issue and Spark Run config cluster selection regression

wezhang:


wezhang:
Addressed #2172, #2167

PR #2195, opened 2018-10-11
Slot UI update: always show but disable the slot panel

SummerSun:
It is not a good practice to show/hide the ui elements when users operate it. We update the behavior to use the enable/disable the elements instead.

comment


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/webapp/webappconfig/ui/WebAppSettingPanel.java
```diff
@@ -542,7 +537,8 @@ public void enableDeploymentSlotPanel() {
 
     private void toggleDeployPanel(boolean isUsingExisting) {
         pnlExist.setVisible(isUsingExisting);
-        pnlDeploySlot.setVisible(isUsingExisting);
+        deployToSlotCheckBox.setVisible(isUsingExisting);
```
Eskibear: It's related with "new webapp" / "existing webapp". Nothing to do with the slots.

SummerSun: Yes, move the pnlSlot to pnlExisting, then we can manage the elements visibility in one entry.

PR #2201, opened 2018-10-12
Make deployment slot area hideable

SummerSun:
#2213 Provide an option that if users don't care about the deployment slot, the area could be hidden.

comment


SummerSun:
test it please

SummerSun:
test this please

Issue #2211, opened 2018-10-15
Add operation to delete a deployment slot


SummerSun:
In the deployment slot right-click operations, to add an operation `Delete` to allow users to delete a deployment slot from the toolkits conveniently.

PR #2216, opened 2018-10-15
Add operation to delete a deployment slot

SummerSun:
#2211 Add the `Delete` operation in the context menu of deployment slots to delete a deployment slot.


Utils/azuretools-core/src/com/microsoft/azuretools/core/mvp/model/webapp/AzureWebAppMvpModel.java
```diff
@@ -418,6 +418,12 @@ public void swapSlotWithProduction(final String subscriptionId, final String app
         slot.swap("production");
     }
 
+    public void deleteDeploymentSlotNode(final String subscriptionId, final String appId,
+                                         final String slotName) throws IOException {
+        final WebApp app = AuthMethodManager.getInstance().getAzureClient(subscriptionId).webApps().getById(appId);
+        app.deploymentSlots().deleteByName(slotName);
```
Eskibear: check whether there will be potential NPE for `AuthMethodManager.getInstance().getAzureClient(subscriptionId).webApps().getById(appId)`
If so, suggest to perform NPE check over all these methods.

SummerSun: This code snippet appears in many places, I will check.

PR #2218, opened 2018-10-15
Handle exception when deleting web app

SummerSun:
A `CloudException` might occur when deleting a web app, we should handle this kind of exception too. Update the caught exception type to generic `Exception` to handle the exceptions during deleting a web app.


Eskibear:
Should catch `CloudException` rather than all exceptions.

SummerSun:
I think any exception occurs, we may inform the users because it will reflect on the UI that the node is still there. Otherwise, it makes users confusing. In other words, what kind of exception do you think that we don't need to show?

PR #2224, opened 2018-10-17
Re-enable cache in Travis CI

Eskibear:
1. Re-enable cache.
2. Limit to build only master/release/develop branch. 
3. In non-PR build, build Eclipse and IntelliJ plugins together.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,5 @@ branches:
   - release
   - develop
 script:
-- travis_wait 40 ./BuildDevint -q -e eclipse
-script:
 - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then travis_wait 40 ./BuildDevint -q -e eclipse; fi'
 - 'if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then travis_wait 40 ./BuildDevint -q; fi'
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,25 @@ language: java
 sudo: false
 jdk:
 - oraclejdk8
+before_cache:
+  - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,7 +2,23 @@ language: java
...
 before_install:
 - export MAVEN_OPTS="-Xmx768M"
+branches:
+  only:
+  - master
+  - release
+  - develop
 script:
-- travis_wait 80 ./BuildDevint -q -e eclipse
+- 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then travis_wait 40 ./BuildDevint -q -e eclipse; fi'
```
wezhang: Can `if  ; then ... else ...  fi` with the following one?

Eskibear: I myself agree with you. But I don't know whether there any trick inside, so I just follow the [official docs](https://docs.travis-ci.com/user/pull-requests) 

I can have a try.

Eskibear:
@wezhang Yes, the difference is whether to build Eclipse plugin. Just the same as what we did in Jenkins. 
a) in non-PR build, we build Eclipse for robustness. 
b) in PR build, we disable it for efficiency. 
Do you have any concern?

.travis.yml
Eskibear: Oh, now I know why they write so in the docs. It's more clear than a single line script. So I tend not to change it now.

Eskibear:
> Are you going to clean up the cache for develop branch building?

a). The best solution I think is to use cache only in PR builds. But I don't see simple and graceful config for that.
b). If we merge this now as it is, I'm afraid I've got to regularly clean up develop branch cache, either manually or by a bot. 

PS: Cache in develop branch has one advantage: when a PR is created and built for the first time, it loads cache in develop branch (if exists).

wezhang:
NP, go ahead!

PR #2227, opened 2018-10-18
Add deployment slot property view editor creator and interface

SummerSun:


SummerSun:
Create `WebAppBasePropertyViewProvider` and inherits from it to create new provider to `DeploymentSlotPropertyViewProvider`. And include the change list in a unified PR to make it more easy to track.

Issue #2158, opened 2018-10-09
[IntelliJ]The  log will not show "job run successfully" when in Link Livy Cluster dialog, do not set “Yarn Endpoint”


v-xiaocw:
Build: azure-toolkit-for-intellij-2018.1.PR.2202.10-08-2018
Repro Steps:
1.Right click on HDInsight ,select "Link A Livy Cluster"
2.Click the dropdown list of the "cluster info,"then select "Livy service",set up other items and leave "yarn endpoint" vacant.
3.Submit job to the linked cluster with the settings above 
Result:
![image](https://user-images.githubusercontent.com/43333874/46663437-edbb8580-cb72-11e8-94c0-81c5298ecc4b.png)
Expect:
Should show "Job run successfully" like below
![image](https://user-images.githubusercontent.com/43333874/46663586-5a368480-cb73-11e8-9ff5-12dcb9e3e8dc.png)

v-jiche:
Still Repro.
Build: azure-toolkit-for-intellij-2018.2.71.10-31-2018

jingyanjingyan:
Repro with develop 869

jingyanjingyan:
Consider make the log more friendly.

lcadzy:
After check,we find out once user dosen't fill the yarn endpoint when linking the livy cluster,we can't get the job log and result since these are based on the endpoint.

Issue #2085, opened 2018-09-25
[IntelliJ][HDP3.0][Regression] Submit success, but in log show UNDEFINED status


jingyanjingyan:
Build: PR2087

Repro Steps:
1. Submit sample.LogQuery to a HDP3.0 cluster "spark231HDI40-reg"

Result:
Success but get log:

 Package and deploy the job to Spark cluster
INFO: Begin uploading file C:\Users\v-yajing\IdeaProjects\maventest1\out\artifacts\maventest1_DefaultArtifact\default_artifact.jar to Azure Blob Storage Account wasbs://spark231hdi40-reg-2018-09-25t00-57-59-798z@jillstormtest.blob.core.windows.net/SparkSubmission/2018/09/25/696d80cb-0210-4f69-95f2-735f1b2445e8/default_artifact.jar ...
INFO: Submit file to azure blob 'wasbs://spark231hdi40-reg-2018-09-25t00-57-59-798z@jillstormtest.blob.core.windows.net/SparkSubmission/2018/09/25/696d80cb-0210-4f69-95f2-735f1b2445e8/default_artifact.jar' successfully.
LOG: Warning: Master yarn-cluster is deprecated since 2.0. Please use master "yarn" with specified deploy mode instead.
LOG: 18/09/25 06:30:08 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
LOG: Warning: Skip remote jar wasbs://spark231hdi40-reg-2018-09-25t00-57-59-798z@jillstormtest.blob.core.windows.net/SparkSubmission/2018/09/25/696d80cb-0210-4f69-95f2-735f1b2445e8/default_artifact.jar.
LOG: 18/09/25 06:30:09 INFO AzureIaasSink: Init starting. Initializing MdsLogger.
LOG: 18/09/25 06:30:09 INFO AzureIaasSink: Init completed.
LOG: 18/09/25 06:30:09 INFO WasbAzureIaasSink: Init completed.
LOG: 18/09/25 06:30:09 INFO MetricsSinkAdapter: Sink azurefs2 started
LOG: 18/09/25 06:30:09 INFO MetricsSystemImpl: Scheduled Metric snapshot period at 60 second(s).
LOG: 18/09/25 06:30:09 INFO MetricsSystemImpl: azure-file-system metrics system started
[...]

PR #2255, opened 2018-10-24
Add ignore unknown property setting in case converting AppAttempt object failure

lcadzy:


lcadzy:
Address #2085

lcadzy:
Address #2158

Issue #2261, opened 2018-10-25
 Can't show the multiple slots properties under the same webapp


lxhmavis:
Build: Develop819
OS: Windows/Mac/Linux

Steps:

1. Deploy a jar file to a new windows webapp, e.g. webapp1.
2. Deploy to slot-1 slot.
3. Deploy to slot-2 slot.
4. Right-click slot-1 to show Properties.
5. Right-click slot-2 to show Properties.

Result:
Plugin can't show the slot-2 properties. 
But we can open multiple slots when they from different webapps.

![image](https://user-images.githubusercontent.com/41565966/47490961-5c3f5980-d87c-11e8-843a-6ce5d6e080ab.png)

SummerSun:
We used the web app id as the resource id for deployment slots. We need the web app id to look for the deployment slot. But all deployment slots share one web app id, that causes the resource is treated existing and that further causes the problem.

Will change to use the deployment id as the unique resource id and write the web app id as user data into the virtual file to fix it.

bsaby:
Test passed.
Build used: daily557

PR #2266, opened 2018-10-26
Put web app id in user data when creating deployment slot properties virtual file

SummerSun:
Fix bug #2261


SummerSun:
test this please

Issue #2265, opened 2018-10-25
Intellij Idea Azure plugin broken, won't sign in, gets the application stuck


polarapfel:
The Azure Intellij Idea plugin seems to be broken.

It will not sign into Azure. When choosing either sign-in method, the sign-in proceeds results in an endless loop, with no ability to cancel the dialog.

It's stuck here: https://developercommunity.visualstudio.com/storage/attachments/45350-annotation.png

I tried on the latest Idea release (full version) on two Windows machines and a Ubuntu 18.04 setup.

This should be a P0 issue. Please let me know what you need from me to help get to the root cause.

thanks,

Tobias Weisserth

Eskibear:
It should pop up a new dialog to let you sign in. Sometimes we notice the dialog does pop up, but it's in the back and is blocked by the main dialog, so you might not see it. I'm not sure if it's your case, can you please have a check? IntelliJ IDEA's API and its UI behavior change very often, do you mean **2018.2 ultimate** by `latest Idea release`.

polarapfel:
Hey,

I am running Idea Ultimate, release 2018.2.5. There is no dialog popping up, neither in the foreground, or background.

If there was a dialog in the background, that's part of Idea's windows, I couldn't switch to it anyway, as the Sign In Progress window locks the focus. There isn't even a way to cancel or close it to get back to Idea. I need to kill Idea from the task manager or process list and restart it, to get back to my workspace (and that ought to be filed as a bug as well I guess).

Please let me know how I can help you debug this. I am eager to get this resolved, as I've been looking forward to having Azure integration in Idea.

thanks,

Tobias Weisserth

Eskibear:
@polarapfel 2018.2.5 was released after we released the plugin, that's how the regression comes. We are fixing it. 

And if you want to try this plugin now, as previous versions were covered in our test, you can use a previous version of the IDEA (e.g. [IntelliJ IDEA 2018.2.3](https://download.jetbrains.com/idea/ideaIU-2018.2.3.exe) )

Eskibear:
It also occurs in 2018.3, see #2279

polarapfel:
Fixed in latest release. Just tested and works. I'll re-test on my other setups and will let you know if there is still an issue. Thanks!

Flanker32:
Close as the issue has been fixed

PR #2305, opened 2018-10-30
Set timeout for JavaFX panel init.

Eskibear:
A workaround for #2265 . 

Root cause:
Potential deadlock in JavaFX panel intialization. 

This PR:
Set a 5-second timeout for the initialization, so the window of IDEA won't be frozen. 

In the future, we'll use AAD Device Login to completely resolve this issue.


Eskibear:
A similar issue is found here: https://bugs.openjdk.java.net/browse/JDK-8172554
And in the official docs of [SecondaryLoop](https://docs.oracle.com/javase/8/docs/api/java/awt/SecondaryLoop.html), the sample has the same code logic. 
```java
   SecondaryLoop loop;

   JButton jButton = new JButton("Button");
   jButton.addActionListener(new ActionListener() {
        @Override
       public void actionPerformed(ActionEvent e) {
           Toolkit tk = Toolkit.getDefaultToolkit();
           EventQueue eq = tk.getSystemEventQueue();
           loop = eq.createSecondaryLoop();

           // Spawn a new thread to do the work
           Thread worker = new WorkerThread();
           worker.start();

           // Enter the loop to block the current event
           // handler, but leave UI responsive
           if (!loop.enter()) {
               // Report an error
           }
       }
   });

   class WorkerThread extends Thread {
        @Override
       public void run() {
           // Perform calculations
           doSomethingUseful();

           // Exit the loop
           loop.exit();
       }
   }
```

PR #2344, opened 2018-11-09
Introduce latest serverless spark REST API

t-rufang:
- You can check the swagger spec [here](https://github.com/Azure/azure-rest-api-specs-pr/blob/AzureDataLakeAnalytics-Private-Permanent/specification/datalake-analytics/data-plane/Microsoft.DataLakeAnalytics/preview/2018-02-01-preview/activities.json)


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/rest/azure/activities/models/AnalyticsActivity.java
```diff
@@ -0,0 +1,272 @@
+/**
```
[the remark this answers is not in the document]
t-rufang: Resolved in latest commit

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/rest/azure/serverless/spark/models/CreateSparkBatchJobParameters.java
```diff
@@ -29,6 +29,7 @@
 
 /**
  * Parameters used to submit a new Data Lake Analytics spark batch job request.
+ * Some fields are commented since they already exists in the base class SparkSubmissionParameter
```
wezhang: `To leverage Livy(HDInsight) Spark Batch Job codes, some fields are commented since they can be inherited from the base class SparkSubmissionParameter`

wezhang:
@skhushalani, could you help to review the changes with unknown properties ignored?

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/rest/azure/serverless/spark/models/AnalyticsActivity.java
```diff
@@ -145,27 +179,65 @@ public String submitter() {
...
+    public String activityState() {
+        return this.activityState;
+    }
+
+    /**
+     * Get result of the activity. Possible values include: 'None', 'Succeeded', 'Cancelled', 'Failed'.
+     *
+     * @return the result value
+     */
+    public ActivityResult result() {
```
skhushalani: The property is called just result in the response, we should keep it the same for consistency instead of ActivityResult

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/rest/azure/serverless/spark/models/CreateSparkBatchJobParameters.java
```diff
@@ -0,0 +1,440 @@
+/**
```
skhushalani: These classes "seem" autogenerated,
1) if they are autogenerated from autorest, we should add either a header telling that its generated or the folder in which it is placed should be named generated.

2) if they are not autogenerated, can you  reply as to why they are not? does autorest cause any problem?

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/rest/azure/serverless/spark/models/CreateSparkBatchJobParameters.java
```diff
@@ -0,0 +1,440 @@
...
+ */
+
+package com.microsoft.azure.hdinsight.sdk.rest.azure.serverless.spark.models;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.microsoft.azure.hdinsight.spark.common.SparkSubmissionParameter;
+
+/**
+ * To leverage Livy(HDInsight) Spark Batch Job codes, some fields are commented
```
skhushalani: I think we should not couple with SparkSubmissionParameter, if we change the serverless request definition we should do it independent of SparkSubmissionParameter... already we have diverged, we have removed "queue" and a few other properties defined in the livy spec

skhushalani:
I have added my comments. Some minor feedback but should be ok for now.

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/rest/azure/serverless/spark/models/CreateSparkBatchJobParameters.java
t-rufang: The API is auto-generated first and then modified manually. We did some modification since:
1. `autorest` won't add `@JsonIgnoreProperties(ignoreUnknown = true)` for us but we need this feature
2. Some serializaion code need to be added to leverage existing code. e.g. Some class in the API need to implement interfact `IConvertible` to help serialize an object to XML-format or Json-format.

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/rest/azure/serverless/spark/models/CreateSparkBatchJobParameters.java
t-rufang: In fact we don't reply on `queue` or other fields either. Here class `CreateSparkBatchJobParameters` inherit from `SparkSubmissionParameter` will help to leverage exisiting code.

Issue #2364, opened 2018-11-19
Eclipse build failure


Eskibear:
Develop branch build failure after merging PR #2350 . As we've changed Azure SDK version, it will block the dependency version comparison, thus fail the coming PR build. Please take a look.

Jenkins log
```
+ mvn clean install -f /jenkinshome/jenkins/workspace/Azure_Toolkits_for_Java/AzureToolsManager_develop/PluginsAndFeatures/azure-toolkit-for-eclipse/pom.xml -q
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:0.25.0:compile (default-compile) on project com.microsoft.azuretools.core: Compilation failure: Compilation failure:
[ERROR] /jenkinshome/jenkins/workspace/Azure_Toolkits_for_Java/AzureToolsManager_develop/PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/ui/UIFactory.java:[13]
[ERROR] public class UIFactory implements IUIFactory {
[ERROR] ^^^^^^^^^
[ERROR] The type UIFactory must implement the inherited abstract method IUIFactory.getDeviceLoginUI()
[ERROR] /jenkinshome/jenkins/workspace/Azure_Toolkits_for_Java/AzureToolsManager_develop/PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/ui/SignInDialog.java:[276]
[ERROR] } catch (IOException | InvocationTargetException | InterruptedException ex) {
[ERROR] ^^^^^^^^^^^
[ERROR] Unreachable catch block for IOException. This exception is never thrown from the try statement body
[ERROR] /jenkinshome/jenkins/workspace/Azure_Toolkits_for_Java/AzureToolsManager_develop/PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/ui/SignInDialog.java:[312]
[ERROR] } catch (IOException ex) {
[ERROR] ^^^^^^^^^^^
[ERROR] Unreachable catch block for IOException. This exception is never thrown from the try statement body
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[...]

PR #2365, opened 2018-11-19
Fix eclipse build error

SummerSun:
To address issue #2364


PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/ui/SignInDialog.java
```diff
@@ -306,13 +306,8 @@ public void run() {
     }
 
     private void doSignOut() {
-        try {
```
SummerSun: There is no `IOException` will be thrown.

Eskibear:
So the root cause seems to be the change of IOException to ExceptionInInitializerError?

SummerSun:
@Eskibear  Updated the root cause in the description.

PR #2371, opened 2018-11-20
Remove unnecessary properties of AuthContext

SummerSun:
The `webUi` and `redirectUrl` are strongly related to the interactive login, they are not necessarily the properties of the `AuthContext`. Remove them as properties and update the API to have them pass in as parameters.


Utils/azuretools-core/src/com/microsoft/azuretools/adauth/AuthContext.java
```diff
@@ -97,13 +86,17 @@ public String getClientId() {
      * Get token from last authentication result
      *
      * @param lastResult last authentication result
+     * @param webUi web ui
+     * @param redirectUrl redirect url
      * @return authentication result with updated tokens
      * @throws AuthException exception during getting token
      */
-    public AuthResult acquireToken(@NotNull AuthResult lastResult) throws AuthException {
+    public AuthResult acquireToken(@NotNull final AuthResult lastResult, final IWebUi webUi,
```
wezhang: If the `webUi` and `redirectUrl` are only used for Device login, can they be put into DCAuthManager? such as constructor of DCAuthManager

SummerSun: They are only used in interactive login.

SummerSun:
test this please

PR #2382, opened 2018-11-21
Correct pricing tiers

SummerSun:
The latest Premium pricing tiers are P1V2, P2V2 and P3V2. We add our workaround to bypass the SDK to fix it in our code repo and can remove the workaround once SDK fix it.


SummerSun:
Track this in #2385

PR #2388, opened 2018-11-21
Fix render wildfly web container in UI issue

SummerSun:
Move the Linux runtime stack related logic to one place then we won't miss it when any update next time.

### DIFF
--- a/Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
+++ b/Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
@@ -28,6 +28,7 @@ import com.microsoft.azure.management.appservice.AppServicePlan;
 import com.microsoft.azure.management.appservice.JavaVersion;
 import com.microsoft.azure.management.appservice.OperatingSystem;
 import com.microsoft.azure.management.appservice.PublishingProfile;
+import com.microsoft.azure.management.appservice.RuntimeStack;
 import com.microsoft.azure.management.appservice.WebApp;
 import com.microsoft.azure.management.appservice.WebContainer;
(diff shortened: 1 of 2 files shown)


Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
```diff
@@ -417,4 +425,10 @@ public static String getJavaRuntime(@NotNull final WebApp webApp) {
                 return DEFAULT_VALUE_WHEN_VERSION_INVALID;
         }
     }
+
+    public static List<RuntimeStack> getAllJavaLinuxRuntimeStacks() {
+        return Arrays.asList(new RuntimeStack[]{
+            RuntimeStack.TOMCAT_8_5_JRE8, RuntimeStack.TOMCAT_9_0_JRE8,
```
Eskibear: split into 4 lines.

Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
```diff
@@ -406,7 +411,10 @@ public static String getJavaRuntime(@NotNull final WebApp webApp) {
                 webContainer = versions[0].toLowerCase();
                 final String webContainerVersion = versions[1];
                 final String jreVersion = versions[2];
-                if (webContainer.contains("tomcat")) {
+                final List<String> allJavaLinuxWebContainers = getAllJavaLinuxRuntimeStacks().stream()
+                    .filter(r -> r != RuntimeStack.JAVA_8_JRE8)
```
Eskibear: something like `r.stack() != Java` would be more robust

PR #2391, opened 2018-11-21
Fix auth error regressions

wezhang:


SummerSun:
What regressions are we fixing here?

wezhang:
To address #2341

PR #2395, opened 2018-11-22
Fix ftp upload retry logic

SummerSun:
We sometimes meet the exceptions when using FTP to upload an artifact to the web app, the retry logic should handle the IOException to do the real retry.


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/webapp/webappconfig/WebAppRunState.java
```diff
@@ -344,26 +344,26 @@ private int uploadFileViaZipDeploy(@NotNull WebAppBase webapp, @NotNull File zip
...
-                                 @NotNull InputStream stream, RunProcessHandler processHandler) throws IOException {
-        boolean success = false;
-        int count = 0;
-        while (!success && ++count < UPLOADING_MAX_TRY) {
+                                @NotNull InputStream stream, RunProcessHandler processHandler) throws IOException {
+        int retry = UPLOADING_MAX_TRY;
+        while (retry > 0) {
+            try {
+                retry -= 1;
+                ftp.storeFile(path, stream);
```
Eskibear: plz handle the case where it returns false

PR #2396, opened 2018-11-23
 Add cancel flag for device login

SummerSun:
Add the `isCancelled` flag while looping the authentication result because `future.cancel(true)` has no guarantee that a task will actually be canceled.

Only when exceptions occurred, there is necessary to wait for a while then retry.


Eskibear:
It looks not thread-safe. 

If you want to access the flag, see the suggested way of doing this in  [com.intellij.openapi.application.Application](https://github.com/JetBrains/intellij-community/blob/master/platform/core-api/src/com/intellij/openapi/application/Application.java#L19-L37)

SummerSun:
@Eskibear 

I found that `AtomicBoolean` could make the flag thread-safe.
 Here what I want to do is that: the `future.canel(true)` could not guarantee that the task be canceled successfully, our while loop keep creating future tasks with this call `authenticationResult =  ctx.acquireTokenByDeviceCode(deviceCode, callback).get();`.

Let's hold this for a while see if there is a better way to achieve this.

Eskibear:
I know your motivation, and from the link I provided above, IntelliJ suggests to use the following function:
```java
  /**
   * Runs the specified read action. Can be called from any thread. The action is executed immediately
   * if no write action is currently running, or blocked until the currently running write action completes.
   *
   * @param action the action to run.
   */
  void runReadAction(@NotNull Runnable action);
```

If you want, I think accessing it within the `action` should work on a primitive boolean flag.

Issue #2425, opened 2018-12-04
[IntelliJ] Autofix for local console does not launched when the SBT project name includes capital letter


jingyanjingyan:
Build: develop 880

Repro Steps:
1. Create a SBT project with capital letter in name
2. Launch local console

Result:
![image](https://user-images.githubusercontent.com/36756024/49428491-2a0d0a00-f7e1-11e8-98e0-621b06cd2114.png)

jingyanjingyan:
Verify with Develop 894, fixed, Autofix diaog pops when launch local console.

Issue #2423, opened 2018-12-04
[IntelliJ]Warning cannot find cluster when there is no main class, run livy interactive console


jingyanjingyan:
Build: Develop 880

Repro Steps:
1. Make sure there is no config file
2. Open a scala file, run livy interactive console

Result:
Warning cannot find the target cluster
![image](https://user-images.githubusercontent.com/36756024/49422672-c201f800-f7cf-11e8-9c1b-1c1bf2675832.png)

The config file include cluster but does not have main class actually
![image](https://user-images.githubusercontent.com/36756024/49422710-e231b700-f7cf-11e8-8bb8-e2b29670d814.png)

jingyanjingyan:
Repro with develop 903

jingyanjingyan:
Expect:
When user click Edit, the cluster should be empty cause user haven't choose any one.
![image](https://user-images.githubusercontent.com/36756024/49858550-2c4e1480-fe30-11e8-9034-217c10b4f05f.png)

jingyanjingyan:
Verify with build develop 914:
Click continue anyway, the livy interactive console does not launch and no useful warning message for it.

v-jiche:
Verified. Fixed.
Build:
azure-toolkit-for-intellij-2018.3.develop.943.12-29-2018

PR #2470, opened 2018-12-11
Refactor Spark Run configuration validations

wezhang:
Now, the run configuration can be saved with errors, but pop up the dialog before running.


wezhang:
Addressed #2423, #2425

Issue #2461, opened 2018-12-07
[IntelliJ][ReportedByUser] Uncaught Exception [Correlation ID: 1aa1fd05-2bac-4b65-b33e-00a1eef26e93] Execution of class com.microsoft.aad.adal4j.AcquireTokenCallable failed. com.microsoft.aad.adal4j.AuthenticationException: AADSTS70016: Pending end-user authorization.


tilumi:
IntelliJ build version: 2018.3.1 IC-183.4588.61
OS: Windows Server 2016
JDK: JetBrains s.r.o 1.8.0_152-release
Plugin version: 3.15.0-2018.3
Additional Info: None
Parent component: 
```
com.intellij.openapi.ui.impl.DialogWrapperPeerImpl$MyDialog$DialogRootPane[,8,31,824x494,invalid,layout=javax.swing.JRootPane$RootLayout,alignmentX=0.0,alignmentY=0.0,border=,flags=449,maximumSize=,minimumSize=,preferredSize=]
```
Error message: 
```
IdeaLoggingEvent[message=[Correlation ID: 1aa1fd05-2bac-4b65-b33e-00a1eef26e93] Execution of class com.microsoft.aad.adal4j.AcquireTokenCallable failed., throwable=com.microsoft.aad.adal4j.AuthenticationException: AADSTS70016: Pending end-user authorization.
Trace ID: be00a398-62f1-41be-931c-bdda48890800
Correlation ID: 1aa1fd05-2bac-4b65-b33e-00a1eef26e93
Timestamp: 2018-12-07 08:36:22Z
	at com.microsoft.aad.adal4j.AdalTokenRequest.executeOAuthRequestAndProcessResponse(AdalTokenRequest.java:117)
	at com.microsoft.aad.adal4j.AuthenticationContext.acquireTokenCommon(AuthenticationContext.java:928)
	at com.microsoft.aad.adal4j.AcquireTokenCallable.execute(AcquireTokenCallable.java:70)
	at com.microsoft.aad.adal4j.AcquireTokenCallable.execute(AcquireTokenCallable.java:38)
	at com.microsoft.aad.adal4j.AdalCallable.call(AdalCallable.java:47)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurren

wezhang:
[...]

Issue #2368, opened 2018-11-20
[IntelliJ] Find error in idea.log when Log in with device code


t-rufang:
1. Put following string into `C:\Users\<username>\AzureToolsForIntelliJ\AadProvider.json`
```
{
    "EnableDeviceLogin": true
}
```
2. Sign in with device code
3. Check the idea.log, we can find the following error log
```
2018-11-20 15:31:26,082 [ 112271]  ERROR - d.adal4j.AuthenticationContext - [Correlation ID: f7233c5b-b241-493e-a720-8bb51f9f5ef2] Execution of class com.microsoft.aad.adal4j.AcquireTokenCallable failed. 
com.microsoft.aad.adal4j.AuthenticationException: AADSTS70016: Pending end-user authorization.
Trace ID: 86c96d1a-5e11-4ed1-abef-585504db0600
Correlation ID: f7233c5b-b241-493e-a720-8bb51f9f5ef2
Timestamp: 2018-11-20 07:31:25Z
	at com.microsoft.aad.adal4j.AdalTokenRequest.executeOAuthRequestAndProcessResponse(AdalTokenRequest.java:117)
	at com.microsoft.aad.adal4j.AuthenticationContext.acquireTokenCommon(AuthenticationContext.java:928)
	at com.microsoft.aad.adal4j.AcquireTokenCallable.execute(AcquireTokenCallable.java:70)
	at com.microsoft.aad.adal4j.AcquireTokenCallable.execute(AcquireTokenCallable.java:38)
	at com.microsoft.aad.adal4j.AdalCallable.call(AdalCallable.java:47)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
2018-11-20 15:31:31,493 [ 117682]  ERROR - d.adal4j.AuthenticationContext - [Correlation ID: f7233c5b-b241-493e-a720-8bb51f9f5ef2] Execution of class com.microsoft.aad.adal4j.AcquireTokenCallable failed. 
com.microsoft.aad.adal4j.AuthenticationException: AADSTS70016: Pending end-user authorization.
Trace ID: 86c96d1a-5e11-4ed1-abef-585534db0600
[...]

PR #2490, opened 2018-12-11
Enable Device Login by default.

Eskibear:
screen shot 2018-12-11 at 9 30 39 pm

I'm not explicitly showing the workflow name "Device Login", which might confuse users. 

#### A known issue
See: #2461 #2368 
We are not able to prevent adal lib from logging error message when it fails each polling, before users complete the login. They use slf4j, and AFAIK we [cannot set level for slf4j logger on the fly](https://stackoverflow.com/a/2622044).


Eskibear:
After clicking "sign in", current UI is as following, let's see if the wording is proper. 

screen shot 2018-12-12 at 9 53 12 am

Issue #2472, opened 2018-12-11
[IntelliJ]Spark Livy interactive session console blocked!


v-xiaocw:
Build: develop 889

intellij:  IntelliJ IDEA 2018.3.1 (Community Edition)
Build #IC-183.4588.27, built on November 28, 2018
JRE: 1.8.0_152-release-1343-b16 amd64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
Windows 10 10.0
 Repro Steps:
1.Create a new project 
2.Click **Run Spark Livy interactive Session Console** on Tools drop-down menu
3.After you close the error that  **cannot find the target cluster**,fill in the configuration file with the main class.
4..Click **Run Spark Livy interactive Session Console** on Tools drop-down menu **again**.
5.Enter SC. or Spark
Result:
After step 4 must wait for more than 2 minutes,then show up as follow picture:
![image](https://user-images.githubusercontent.com/43333874/49785485-91412780-fcd5-11e8-92c8-1dd596f59ea2.png)
The result of step 5:
![image](https://user-images.githubusercontent.com/43333874/49785459-7c649400-fcd5-11e8-83da-fbb3d4eb27fd.png)

jingyanjingyan:
Fix with develop 901

Issue #2468, opened 2018-12-11
[Intellij][2018.3][HDInsight] IntelliJ 183 regresion of Spark livy interactive console


wezhang:
```
Failed to get PID with winp, fallback to default logic

org.jvnet.winp.NotWindowsException: java.lang.NoSuchFieldException: handle
	at org.jvnet.winp.WinProcess.<init>(WinProcess.java:48)
	at com.intellij.execution.process.OSProcessUtil.createWinProcess(OSProcessUtil.java:127)
	at com.intellij.execution.process.OSProcessUtil.getProcessID(OSProcessUtil.java:104)
	at com.intellij.execution.impl.DefaultJavaProgramRunner$AttachDebuggerAction$1.lambda$startNotified$0(DefaultJavaProgramRunner.java:231)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at com.intellij.util.concurrency.SchedulingWrapper$MyScheduledFutureTask.run(SchedulingWrapper.java:228)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.NoSuchFieldException: handle
	at java.lang.Class.getDeclaredField(Class.java:2070)
	at org.jvnet.winp.WinProcess.<init>(WinProcess.java:43)
	... 9 more
```

jingyanjingyan:
Fix with develop 901

PR #2495, opened 2018-12-12
Fix Spark livy console regression of IC 183 win process

wezhang:
IDEA 183 adds an `AttachDebuggerAction` for the process handler inherited `BaseProcessHandler`, which requires windows process. (refer to JetBrains/intellij-community@6cf5b9f890e283d7507954e2b529038dc5c02670)

Fix the issue with inheriting from ProcessHandler.


wezhang:
To address #2468, also help #2472

Issue #2547, opened 2018-12-26
[IntelliJ] Reject aris certificate, warning message need update


jingyanjingyan:
Build: Develop 914

Repro Steps:
1. Link an aris cluster and reject when pops certificate dialog

Result:
Warning message says livy URL
![image](https://user-images.githubusercontent.com/36756024/50436625-d1cdb180-0921-11e9-87b5-9b35d21dbd81.png)

Expect:
Should change to Host?

v-jiche:
Verified, Fixed.
azure-toolkit-for-intellij-2018.3.develop.923.12-27-2018
Change livy URL to host.
![image](https://user-images.githubusercontent.com/35950097/50471442-844a5680-09ac-11e9-8207-9d952a59b40a.png)

PR #2548, opened 2018-12-26
Change err msg when user rejects ca

lcadzy:
since now linking aris dosen't need livy and yarn url ,update the err msg


lcadzy:
#2547

Issue #2567, opened 2018-12-27
[IntelliJ][Linux]Frequent pop-ups when list HDInsight clusters.


v-jiche:
Build:
azure-toolkit-for-intellij-2018.3.develop.923.12-27-2018

Repro Steps:
1. Open Azure.
2. Click HDInsight to list clusters.

Result:
1. Pop-up windows pop up frequently.
![image](https://user-images.githubusercontent.com/35950097/50473366-916b4380-09b4-11e9-95ee-ba89d48d094d.png)
2. Clusters is always refreshing.
![image](https://user-images.githubusercontent.com/35950097/50474772-2ae92400-09ba-11e9-8941-d9338268a6a4.png)

wezhang:
@t-rufang , could you check it with other refresh issue #2392

jingyanjingyan:
java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:903)
	at java.util.ArrayList$Itr.next(ArrayList.java:853)
	at com.microsoft.tooling.msservices.helpers.collections.ObservableList.fireChangeListenerEvent(ObservableList.java:124)
	at com.microsoft.tooling.msservices.helpers.collections.ObservableList.add(ObservableList.java:76)
	at com.google.common.collect.ForwardingList.standardAdd(ForwardingList.java:139)
	at com.microsoft.tooling.msservices.helpers.collections.ObservableList.add(ObservableList.java:44)
	at com.microsoft.tooling.msservices.serviceexplorer.Node.addChildNode(Node.java:193)
	at com.microsoft.azure.hdinsight.serverexplore.hdinsightnode.StorageAccountNode.lambda$refreshItems$1(StorageAccountNode.java:107)
	at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.Iterator.forEachRemaining(Iterator.java:116)
	at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
	at java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151)
[...]

PR #2590, opened 2018-12-29
Fix potential infinity loop issue

t-rufang:
if A() <- B() means method B is refered in method A. We have:
1. In class `RefreshableNode`, `load(boolean forceRefresh)` <- `refreshItems(SettableFuture<List<Node>> future, boolean forceRefresh)` <- `refreshItems()`
2. In class `ClusterNode`, `refreshItems()` <- `load(boolean forceRefresh)`.
Here infinity loop happends.


t-rufang:
Fix #2567

Issue #2488, opened 2018-12-11
[IntelliJ] The edit Auth status will not be saved


jingyanjingyan:
Build: Develop 894

Repro Steps:
1. Create a new HDInsight config file, click "Remotely run Cluster", set main class, and advanced settings, Auth pass now.
2. Re-open the config file, edit the Auth to use wrong password, Auto fails now.
3. Remote debug, click Edit when pops warning says Auth fails.
4. In the config dialog, Auth is fails
![image](https://user-images.githubusercontent.com/36756024/49801444-1e21ca80-fd42-11e8-8528-d6a964f5f474.png)

5. Cancel
6. Open the config file from tool bar, the Auth status is pass
![image](https://user-images.githubusercontent.com/36756024/49801529-56290d80-fd42-11e8-9773-34b348a31f46.png)

7. Remote debug, warning Auth fails again
![image](https://user-images.githubusercontent.com/36756024/49801541-6640ed00-fd42-11e8-8070-495e1f72a86a.png)

Result:
The config file from toolbar is different from edit entry, it makes user confuse in step 7.

v-jiche:
Verified. Fixed.
Build: azure-toolkit-for-intellij-2018.3.develop.953.01-02-2019

Whether you click edit  or open the configuration  from the toolbar, the result is failed.

PR #2605, opened 2019-01-02
Fix the regression Spark AdvConf ssh password loading

wezhang:
`advModel.credentialStoreAccount` depends on `clusterName` property which has to be set before getting `credentialStoreAccount`


wezhang:
Addressed #2488 regression

Issue #2511, opened 2018-12-13
[IntelliJ][Spark on ADL] Should filter storage account for adls cluster


jingyanjingyan:
Build: Private 2675

Repro Steps:
1. Create a config of spark on adl, click on remote conifg
2. Check the combo box of storage list

Result:
All types are listed
![image](https://user-images.githubusercontent.com/36756024/49918157-06cb1480-fedd-11e8-9203-8e78f4a08e4f.png)

Expect:
Filter for adls account?

jingyanjingyan:
Verify with develop 946, still repro.

PR #2610, opened 2019-01-02
Change Serverless Cluster Default Storaget Type

lcadzy:
#2511


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/common/azure/serverless/AzureSparkServerlessCluster.java
```diff
@@ -57,6 +59,7 @@
                                                     DestroyableCluster,
```
wezhang: Could you help to change the Class name to `AzureSparkCosmosCluster`?

Issue #2466, opened 2018-12-10
[Intellij][HDInsight] An exception "ExitCodeException exitCode=10" is classified as a tool error


wezhang:
```
Stack trace: ExitCodeException exitCode=10:
at org.apache.hadoop.util.Shell.runCommand(Shell.java:944)
at org.apache.hadoop.util.Shell.run(Shell.java:848)
at org.apache.hadoop.util.Shell$ShellCommandExecutor.execute(Shell.java:1142)
at org.apache.hadoop.yarn.server.nodemanager.DefaultContainerExecutor.launchContainer(DefaultContainerExecutor.java:237)
at org.apache.hadoop.yarn.server.nodemanager.containermanager.launcher.ContainerLaunch.call(ContainerLaunch.java:317)
at org.apache.hadoop.yarn.server.nodemanager.containermanager.launcher.ContainerLaunch.call(ContainerLaunch.java:83)
at java.util.concurrent.FutureTask.run(FutureTask.java:266)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
at java.lang.Thread.run(Thread.java:748)
Container exited with a non-zero exit code 10
```

wezhang:
advanced info:

```
com.microsoft.azure.hdinsight.spark.common.SparkJobException: The Spark job failed to start due to 	at org.apache.hadoop.yarn.server.nodemanager.containermanager.launcher.ContainerLaunch.call(ContainerLaunch.java:317)
	at org.apache.hadoop.yarn.server.nodemanager.containermanager.launcher.ContainerLaunch.call(ContainerLaunch.java:83)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)


Container exited with a non-zero exit code 10
Failing this attempt. Failing the application.
	at com.microsoft.azure.hdinsight.spark.common.SparkBatchJob.lambda$awaitStarted$58(SparkBatchJob.java:1252)
[...]

Issue #1827, opened 2018-07-20
[IntelliJ][Spark] Spark Job with exception NoSuchClass didn't fetch the Log


wezhang:

lcadzy:
After repo, it shows application can't launch. We classify this as user error.
![image](https://user-images.githubusercontent.com/43339305/50813125-54a63180-1350-11e9-82a4-895db7e290a5.png)

lcadzy:
same with issue #2466

PR #2623, opened 2019-01-08
classify user error for wrong class name

lcadzy:
When user inputs wrong class name , sparkjobexception can happen
image

This pr classifies this exp as user exp
#1827 
#2466 
Since clusters of different versions have different detail error message , we just treat SparkJobException as user exception at present.

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/common/classifiedexception/SparkUserException.kt
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/common/classifiedexception/SparkUserException.kt
@@ -21,6 +21,7 @@
  */
 package com.microsoft.azure.hdinsight.common.classifiedexception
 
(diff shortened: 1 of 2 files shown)


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/common/classifiedexception/SparkUserException.kt
```diff
@@ -21,7 +21,11 @@
  */
 package com.microsoft.azure.hdinsight.common.classifiedexception
 
+import com.microsoft.azure.hdinsight.spark.common.SparkJobException
 import com.microsoft.azure.hdinsight.spark.common.YarnDiagnosticsException
+import org.apache.commons.lang.exception.ExceptionUtils
+
+const val NotLaunchError = "Container exited with a non-zero exit code 10"
```
wezhang: Maybe we needn't to limit the exit code to only 10

lcadzy:
#2659

Issue #2580, opened 2018-12-28
[IntelliJ][ReportedByUser] [Log][Spark]Azure Plugin for IntelliJ Error - [HDInsight] invalid connection string when submit with  azure blob 


v-jiche:
Build: 
azure-toolkit-for-intellij-2018.3.develop.928.12-28-2018

Repro Steps:
1. Link cluster spark22blobhdi
2. Create a HDInsight config file, set cluster spark22blobhdi, set main class, set storage Type to Azure blob and correct storage account ,then **keep the key empty,** save config file
3. Click remote submit, click "Continue anyway"

Result:
![image](https://user-images.githubusercontent.com/35950097/50508389-92dd5000-0a36-11e9-9bcd-8e3b177be08b.png)

Error details:
java.lang.IllegalArgumentException: Invalid connection string.
	at com.microsoft.azure.storage.CloudStorageAccount.parse(CloudStorageAccount.java:270)
	at com.microsoft.tooling.msservices.helpers.azure.sdk.StorageClientSDKManager.getCloudStorageAccount(StorageClientSDKManager.java:742)
	at com.microsoft.tooling.msservices.helpers.azure.sdk.StorageClientSDKManager.getCloudBlobClient(StorageClientSDKManager.java:761)
	at com.microsoft.tooling.msservices.helpers.azure.sdk.StorageClientSDKManager.getBlobContainers(StorageClientSDKManager.java:116)
	at com.microsoft.azure.hdinsight.spark.jobs.JobUtils.getSparkClusterDefaultContainer(JobUtils.java:433)
	at com.microsoft.azure.hdinsight.spark.jobs.JobUtils.uploadFileToAzure(JobUtils.java:454)
	at com.microsoft.azure.hdinsight.spark.jobs.JobUtils.lambda$deployArtifact$9(JobUtils.java:661)
	at rx.internal.operators.OnSubscribeFromCallable.call(OnSubscribeFromCallable.java:48)
	at rx.internal.operators.OnSubscribeFromCallable.call(OnSubscribeFromCallable.java:33)
	at rx.Observable.unsafeSubscribe(Observable.java:10327)
	at rx.internal.operators.OnSubscribeMap.call(OnSubscribeMap.java:48)
	at rx.internal.operators.OnSubscribeMap.call(OnSubscribeMap.java:33)
	at rx.Observable.unsafeSubscribe(Observable.java:10327)
[...]

PR #2666, opened 2019-01-15
Fix multiple storage type events which causes config reload failure

lcadzy:
#2580


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/spark/common/SparkSubmitStorageType.java
```diff
@@ -29,6 +29,7 @@
 import com.microsoft.azuretools.azurecommons.helpers.NotNull;
 
 public enum SparkSubmitStorageType {
+    USERHINT("Please Choose a storagetype or User the default"),
```
lcadzy: will add storage type check in runconfiguration for new type

PR #2673, opened 2019-01-16
Fix setdata async which causes multiple storage event and data loss

lcadzy:


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/spark/ui/SparkSubmissionJobUploadStorageWithUploadPathPanel.kt
```diff
@@ -144,11 +143,15 @@ class SparkSubmissionJobUploadStorageWithUploadPathPanel
                     if (checkEvent.preClusterName != null) {
                         // If preClusterName is not set, the event is triggered by creating config or reloading a saved config
                         model.storageAccountType = clusterDetail.defaultStorageType
+                    } else
+                    {
```
lcadzy: format next iteration.....

Issue #2719, opened 2019-01-25
Wrong font on Azure Device Login dialog


bsaby:
Windows:
![image](https://user-images.githubusercontent.com/22470312/51730805-cfdb4780-20b3-11e9-81af-88689a799507.png)
Mac:
![image](https://user-images.githubusercontent.com/22470312/51730871-0add7b00-20b4-11e9-90e2-541522a98cc5.png)
Linux:
![image](https://user-images.githubusercontent.com/22470312/51730811-d5d12880-20b3-11e9-91b8-816c34739297.png)

bsaby:
Build: pre-release bits
- IntelliJ 18.3.x: Fixed and test passed.
- IntelliJ 18.2.x: **Repro**
![image](https://user-images.githubusercontent.com/22470312/51953695-4e4f3500-2478-11e9-9389-4ccd01d3169b.png)

bsaby:
Build: 3.18.0(pre-release bits)
Fixed and the test passed.

PR #2731, opened 2019-01-28
Apply same font to Device Login Dialog

Flanker32:
Use same font for device login dialog, for issue #2719 
After:
after
Before:
before

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azuretools/ijidea/ui/DeviceLoginWindow.java
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azuretools/ijidea/ui/DeviceLoginWindow.java
@@ -32,11 +32,17 @@ import com.microsoft.aad.adal4j.AuthenticationResult;
 import com.microsoft.aad.adal4j.DeviceCode;
 import com.microsoft.intellij.ui.components.AzureDialogWrapper;
 import org.apache.log4j.Level;
-import org.apache.log4j.LogManager;
 import org.apache.log4j.Logger;
 import org.jetbrains.annotations.Nullable;
 
+import javax.swing.JComponent;
+import javax.swing.JEditorPane;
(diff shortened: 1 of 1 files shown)


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azuretools/ijidea/ui/DeviceLoginWindow.java
```diff
@@ -82,6 +84,12 @@ public DeviceLoginWindow(final AuthenticationContext ctx, final DeviceCode devic
                 }
             }
         });
+        // Apply JLabel's font to JEditorPane
+        final Font font = UIManager.getFont("Label.font");
```
andxu: null checker

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azuretools/ijidea/ui/DeviceLoginWindow.java
```diff
@@ -82,6 +84,14 @@ public DeviceLoginWindow(final AuthenticationContext ctx, final DeviceCode devic
...
+        if (font != null) {
+            final String cssRule = String.format("body { font-family: %s ; font-size: %s pt ;}", font.getFamily(),
+                font.getSize());
+            ((HTMLDocument) editorPanel.getDocument()).getStyleSheet().addRule(cssRule);
```
andxu: I don't see any reason to using css rules here, what do you think of adding in-place css insder createHtmlFormatMessage?

Flanker32: Changed to use JEditorPane.HONOR_DISPLAY_PROPERTIES, which will use component setting when font or foreground color is not specified in the styled text.

Issue #2778, opened 2019-02-27
[Enhancement]The deployed web app list should be sorted alphabetically


sallyur:
### Environment
1. OS: All
2. IDE: Eclipse IDE for Enterprise Java Developers 2018-12
3. JDK: 1.8.0_191
4. Build: 994

### Steps:
1. Make sure at least two subscriptions are selected
1. Right click project and select Azure -> publish as Azure web app

### Result:
This list sorts a bit confusing
![image](https://user-images.githubusercontent.com/28860891/53480320-eeb76a00-3ab5-11e9-8d64-f32884a40f7b.png)

zmssp:
Please verify, it should be fixed

bsaby:
The App service with uppercase letters are in front of apps with lowercase letters, expected?
![image](https://user-images.githubusercontent.com/22470312/53877648-b3ccad80-4044-11e9-816e-3abd4497b002.png)
The behavior is different from azure explorer, because I find that the order of web apps in azure explorer like below;
![image](https://user-images.githubusercontent.com/22470312/53877995-a82db680-4045-11e9-82de-10aa56dd0a77.png)

ZhengJun003:
Can reproduce this issue in azure toolkits for eclipse 3.23.0

vscjavaci:
Track internal issue by AB#1738756

vscjavaci:
Close this issue since it is tracked in Azure Boards.

PR #2783, opened 2019-02-28
Sort deploy web app dialog by web app name

zmssp:


zmssp:
For this issue, https://github.com/Microsoft/azure-tools-for-java/issues/2778
Sort the table by web app name

PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.webapp/src/com/microsoft/azuretools/webapp/ui/WebAppDeployDialog.java
```diff
@@ -2,10 +2,7 @@
 
 import java.io.InputStream;
 import java.net.URL;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
```
Eskibear: Prefer not to use wildcard import, in case of name collision.

Issue #2788, opened 2019-02-28
ACR Explorer always load when opning with empty repository


bsaby:
### Environment:
1. OS: Windows/Mac/Linux
2. Eclipse: 2018-12 (4.10.0)
3. Bits: dev995
3. ACR: An ACR without any repository and check the admin user enabled.

### Steps:
Right-click the ACR and choose Open ACR Explorer.

### Result:
It is always loading.
![image](https://user-images.githubusercontent.com/22470312/53541149-7c956280-3b53-11e9-9003-21ed1f4274ec.png)
Error log:
```
java.lang.NullPointerException
	at com.microsoft.azuretools.azureexplorer.editors.container.ContainerRegistryExplorerEditor.fillList(ContainerRegistryExplorerEditor.java:707)
	at com.microsoft.azuretools.azureexplorer.editors.container.ContainerRegistryExplorerEditor.listRepo(ContainerRegistryExplorerEditor.java:563)
	at com.microsoft.tooling.msservices.serviceexplorer.azure.container.ContainerRegistryPropertyViewPresenter.lambda$null$9(ContainerRegistryPropertyViewPresenter.java:157)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:185)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:3919)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3550)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1173)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1062)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:156)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:636)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:563)
	at org.eclipse.ui.PlatformUI.createAndRunWorkbench(PlatformUI.java:151)
[...]

PR #2798, opened 2019-03-01
- Fix issue 2788

zmssp:
- Fix issue https://github.com/Microsoft/azure-tools-for-java/issues/2788
- Add null check


Eskibear:
Also change the title

PR #2824, opened 2019-03-05
- Do not show "refresh" button in webapp node

zmssp:
We should not show "refresh" button in web app node, since we do not support deploy slot and have nothing to refresh.

wechat image_20190305091916


zmssp:
Should not modify the common code, close this pr

PR #2872, opened 2019-03-12
Move version to parent pom

zmssp:
Every time when we release a new product, we need to modify several pom.xml for bump version. 
This pr tries to move the version to parent pom, the child pom will read properties from parent pom.


PluginsAndFeatures/AddLibrary/AppInsights/com.microsoft.applicationinsights.feature/pom.xml
```diff
@@ -6,7 +6,7 @@
 	<relativePath>../../../azure-toolkit-for-eclipse/pom.xml</relativePath>
     <groupId>com.microsoft.azuretools.wap4ej.build</groupId>
     <artifactId>parent</artifactId>
-    <version>3.12.0-SNAPSHOT</version>
+    <version>${eclipse.version}</version>
```
andxu: azure.toolkit.eclipse.version

zmssp: use "azuretool.eclipse.version", since the project has already use like "azuretool.sdk.version"

PR #2898, opened 2019-03-21
Use device login as default login method and hide interactive login

Flanker32:
Use device login as default login method and hide interactive login

ScreenShot:
image


Eskibear:
Just to mention that the screenshot is not matching your code change. 
In your code changes, it's "Service Principal", which is what we want. But in the screenshot, the letter "p" is in lower case, which looks strange.

Flanker32:
> Just to mention that the screenshot is not matching your code change.
> In your code changes, it's "Service Principal", which is what we want. But in the screenshot, the letter "p" is in lower case, which looks strange.

It should be "Service Principal", forgot to re-screenshot after fix the case

PR #2902, opened 2019-03-22
Eclipse toolkit support deployment slot

zmssp:
This pr is only for create and deploy to slot. 
The azure explorer slot viewer will commit in another pr.

image


Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
```diff
@@ -499,6 +503,17 @@ public static String getJavaRuntime(@NotNull final WebApp webApp) {
         }
     }
 
+    public static List<DeploymentSlot> getDeployments(WebApp webApp) {
+        if (webApp == null || webApp.deploymentSlots() == null || webApp.deploymentSlots().list() == null) {
+            return new ArrayList<>();
+        }
+        List<DeploymentSlot> result = new ArrayList<>();
```
Flanker32: change to `return new ArrayList<>(webApp.deploymentSlots().list());` ?

PR #2907, opened 2019-03-25
Azure explorer support deployment slot

zmssp:
Deployment slot and webapp will share the same property view.

image


PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.azureexplorer/src/com/microsoft/azuretools/azureexplorer/editors/webapp/DeploymentSlotPropertyEditorInput.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null) {
+            return false;
+        }
```
andxu: bad impl of equals with base class

zmssp: Modify the implementation of equals, and add the hashcode method

Issue #2874, opened 2019-03-12
Telemetry consistent for azure toolkit


zmssp:
This issue is for telemetry consistent, we will use the same schema as vscode.

The full event name = ${platformName} + "/" +  {eventtype}
- platformName: {"AzurePlugin.Eclipse", "AzurePlugin.Intellij"}
- eventtype: {opStart, opEnd, step, info, warn, error}

Example: AzurePlugin.Eclipse/opEnd,  AzurePlugin.Intellij/info

We also define the service name and operation name.
Service name and operation name will put in properties.

The typical service name is: webapp, docker, hdinsight....
The typical operation name is: create-webapp...

zmssp:
https://github.com/Microsoft/azure-tools-for-java/pull/2875#issuecomment-472321854

The problem is how we interact one operation with another. For example, when we deploy the web app, we may first create a new one, then deploy, can we link this two operations?

zmssp:
In order to support transaction trace. 
- Each time when you start a transaction, we will generate a operationId(the operationId will store in thread local), then you can send info,warn,error with this operationId, you need to end this transaction.
- If you start a new transaction without close the previous one, we just generate a new operationId.
- If you send info, error, warn or end transaction without start a transaction, we just give you a new operationId.
- Take care, each time when you start a transaction you need to end it. Or you cannot correctly trace the operation.

The interface design is:

```
public interface Producer {

    void startTransaction(String eventName, String operName, Map<String, String> properties);

    void endTransaction(String eventName, String operName, Map<String, String> properties, long time);

    void sendError(String eventName, String operName, ErrorType errorType, String errMsg,
        Map<String, String> properties);

    void sendInfo(String eventName, String operName, Map<String, String> properties);

    void sendWarn(String eventName, String operName, Map<String, String> properties);

}
```

The usage code like:

```
try {
      startTransaction();
      doSomething();
      sendInfo();
} catch (Exception e) {
      sendError();
} finally {
      endTransaction();
}
```

vscjavaci:
Track internal issue by AB#1738828

PR #2935, opened 2019-03-26
Redesign the telemetry interface

zmssp:
* Redesign telemetry interface.
* Provide functional interface, it will make code clear.
* User will use EventUtil to send telemetry, and use operation to control the request tracing.
For more detail, please refer to #2874


Utils/azuretools-core/src/com/microsoft/azuretools/telemetrywrapper/EventUtil.java
```diff
@@ -28,29 +28,107 @@
...
+            logError(operation, ErrorType.userError, e, properties, metrics);
+            if (errorHandle != null) {
+                errorHandle.accept(e);
+            }
+        } finally {
+            operation.complete();
+        }
+    }
+
+    public static <R> R logCommand(String eventName, String operName, Map<String, String> properties,
```
andxu: logCommand is not applicable in this case, since you have executed the method.

zmssp: Just rename to executeWithLog

Issue #2943, opened 2019-03-27
Failed to sign in with device login method


bsaby:
### Environment:
OS: Win/Mac/Linux
JDK: 1.8
Bits: develop1062
### Repro steps:

1. Login Azure using Service Principal method
2. Then Cancel the login process
3. Try Device Flow in the Sign In Error Dialog
![image](https://user-images.githubusercontent.com/22470312/55058732-d6803e00-50a7-11e9-8c28-f446528fb7d4.png)

### Result:
Login fails
![image](https://user-images.githubusercontent.com/22470312/55058691-a8026300-50a7-11e9-9c0f-e8da923f467d.png)
Log:
```
com.microsoft.azuretools.adauth.AuthException: No subscription found in the account
	at com.microsoft.azuretools.authmanage.SubscriptionManager.doSetSubscriptionDetails(SubscriptionManager.java:122)
	at com.microsoft.azuretools.authmanage.SubscriptionManager.updateSubscriptionDetailsIfNull(SubscriptionManager.java:92)
	at com.microsoft.azuretools.authmanage.SubscriptionManager.getSubscriptionDetails(SubscriptionManager.java:76)
	at com.microsoft.tooling.msservices.serviceexplorer.azure.AzureModule.composeName(AzureModule.java:105)
	at com.microsoft.tooling.msservices.serviceexplorer.azure.AzureModule.handleSubscriptionChange(AzureModule.java:220)
	at com.microsoft.tooling.msservices.serviceexplorer.azure.AzureModule.access$100(AzureModule.java:47)
	at com.microsoft.tooling.msservices.serviceexplorer.azure.AzureModule$SignInOutListener.run(AzureModule.java:229)
	at com.microsoft.azuretools.authmanage.AuthMethodManager.notifySignInEventListener(AuthMethodManager.java:97)
	at com.microsoft.azuretools.ijidea.actions.AzureSignInAction.doSignIn(AzureSignInAction.java:133)
	at com.microsoft.azuretools.ijidea.actions.AzureSignInAction.onAzureSignIn(AzureSignInAction.java:117)
	at com.microsoft.intellij.serviceexplorer.azure.SignInOutAction$1.actionPerformed(SignInOutAction.java:46)
	at com.microsoft.tooling.msservices.serviceexplorer.NodeActionListener.actionPerformedAsync(NodeActionListener.java:72)
[...]

Issue #2928, opened 2019-03-26
Fail to initialize javaFX panel in IntelliJ x86 version


bsaby:
### Environment:
OS: Window10
JDK: 1.8
IntelliJ version:
```
IntelliJ IDEA 2019.1 (Ultimate Edition)
	Build #IU-191.6183.62, built on March 21, 2019
	Licensed to IntelliJ IDEA Evaluator
	Expiration date: April 25, 2019
	JRE: 1.8.0_202-release-1483-b39 x86
	JVM: OpenJDK Server VM by JetBrains s.r.o
	Windows 10 10.0

IntelliJ IDEA 2019.1 (Community Edition)
	Build #IC-191.6183.62, built on March 21, 2019
	JRE: 1.8.0_202-release-1483-b39 x86
	JVM: OpenJDK Server VM by JetBrains s.r.o

```

### Repro steps:

1. Run IntelliJ x86 version and install Azure Toolkit for IntelliJ plugin
2. Login Azure using `Service Principal` -> `New` method

### Result:
If local jdk version is 1.8 x64
![image](https://user-images.githubusercontent.com/22470312/54983315-80e45c80-4fe7-11e9-9193-8b53bc0f61a1.png)

If local jdk version is 1.8 x86
![image](https://user-images.githubusercontent.com/22470312/54983372-a5d8cf80-4fe7-11e9-8fb0-0cfe05ae9500.png)

bsaby:
Bits: 3.20.0
The test passed.

Now, the plugin will use device login as the default auth method if users select` Service Principal` -> `New `method and the JavaFX panel will not appear.

Issue #2893, opened 2019-03-20
[IntelliJ][ReportedByUser] Uncaught Exception  java.lang.NullPointerException


konebra:
IntelliJ build version: 2018.3.5 IC-183.5912.21
OS: Mac OS X
JDK: JetBrains s.r.o 1.8.0_152-release
Plugin version: 3.19.0-2018.3
Additional Info: None
Parent component: 
```
com.intellij.openapi.wm.impl.IdeRootPane[,0,22,1280x755,layout=javax.swing.JRootPane$RootLayout,alignmentX=0.0,alignmentY=0.0,border=,flags=449,maximumSize=,minimumSize=,preferredSize=]
```
Error message: 
```
IdeaLoggingEvent[message=null, throwable=java.lang.NullPointerException
	at com.intellij.openapi.wm.impl.InternalDecorator.setTitleActions(InternalDecorator.java:316)
	at com.intellij.openapi.wm.impl.ToolWindowImpl.setTitleActions(ToolWindowImpl.java:309)
	at com.microsoft.intellij.components.ServerExplorerToolWindowFactory.lambda$addToolbarItems$3(ServerExplorerToolWindowFactory.java:374)
	at com.microsoft.azuretools.authmanage.AuthMethodManager.notifySignInEventListener(AuthMethodManager.java:97)
	at com.microsoft.azuretools.ijidea.actions.AzureSignInAction.doSignIn(AzureSignInAction.java:133)
	at com.microsoft.azuretools.ijidea.actions.AzureSignInAction.onAzureSignIn(AzureSignInAction.java:117)
	at com.microsoft.intellij.serviceexplorer.azure.SignInOutAction$1.actionPerformed(SignInOutAction.java:46)
	at com.microsoft.tooling.msservices.serviceexplorer.NodeActionListener.actionPerformedAsync(NodeActionListener.java:72)
	at com.microsoft.tooling.msservices.serviceexplorer.No

wezhang:
@Flanker32 , could you help follow it!

Flanker32:
@konebra  Thanks for your report, could you please share me which method did you use to login? And is there any method could reproduce this issue?

Flanker32:
We didn't remove the tool window listener when close a project, fixed in #3431

Issue #2757, opened 2019-02-16
[IntelliJ][ReportedByUser] Uncaught Exception  java.lang.NullPointerException when sign in/out Azure account


codeclan04:
IntelliJ build version: 2018.3.3 IU-183.5153.38
OS: Mac OS X
JDK: JetBrains s.r.o 1.8.0_152-release
Plugin version: 3.18.0-2018.3
Additional Info: None
Parent component: 
```
com.intellij.openapi.wm.impl.IdeRootPane[,0,22,1440x855,invalid,layout=javax.swing.JRootPane$RootLayout,alignmentX=0.0,alignmentY=0.0,border=,flags=449,maximumSize=,minimumSize=,preferredSize=]
```
Error message: 
```
IdeaLoggingEvent[message=null, throwable=java.lang.NullPointerException
	at com.intellij.openapi.wm.impl.InternalDecorator.setTitleActions(InternalDecorator.java:316)
	at com.intellij.openapi.wm.impl.ToolWindowImpl.setTitleActions(ToolWindowImpl.java:309)
	at com.microsoft.intellij.components.ServerExplorerToolWindowFactory.lambda$addToolbarItems$3(ServerExplorerToolWindowFactory.java:374)
	at com.microsoft.azuretools.authmanage.AuthMethodManager.notifySignOutEventListener(AuthMethodManager.java:106)
	at com.microsoft.azuretools.authmanage.AuthMethodManager.signOut(AuthMethodManager.java:167)
	at com.microsoft.azuretools.ijidea.actions.AzureSignInAction.onAzureSignIn(AzureSignInAction.java:114)
	at com.microsoft.intellij.serviceexplorer.azure.SignInOutAction$1.actionPerformed(SignInOutAction.java:46)
	at com.microsoft.tooling.msservices.serviceexplorer.NodeActionListener.actionPerformedAsync(NodeActionListener.java:72)
	at com.microsoft.tooling.msservices.serviceex

wezhang:
@andxu , could you help follow it? Thanks!

wezhang:
@Flanker32

Flanker32:
@codeclan04 Thanks for your report, this seems caused by wrong behavior when create new service principal with device login, this will be fixed in 3.20.0.

Flanker32:
We didn't remove the tool window listener when close a project, fixed in #3431

PR #2946, opened 2019-03-27
Use device login as the auth method for create service principal

Flanker32:
Use device login as the auth method for create service principal, for issue #2943 #2928 #2893 #2757


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azuretools/ijidea/ui/SignInWindow.java
```diff
@@ -422,7 +422,8 @@ public void doCancelAction() {
     @Override
     public void doHelpAction() {
         JXHyperlink helpLink = new JXHyperlink();
-        helpLink.setURI(URI.create("https://docs.microsoft.com/en-us/azure/azure-toolkit-for-intellij-sign-in-instructions"));
+        helpLink.setURI(URI.create("https://docs.microsoft.com/en-us/azure/azure-toolkit-for-intellij-sign-in" +
```
Eskibear: do not break URL.

Issue #2958, opened 2019-03-29
[Intellij][SparkOnCosmos] Resource Pool Delete Experience


wezhang:
Some customers are reporting that they aren't able to delete spark resource pools. This maybe expected based on our recent deployment. However, they arent seeing an error message or traceId which we can use to determine what the actual issue is. Can you take a look as to how we can show the error message and at the very least the traceId.

![image](https://user-images.githubusercontent.com/7385667/55211353-95626800-5226-11e9-81d8-a408718926ad.png)

jingyanjingyan:
With the released build IntelliJ IDEA 2019.1 RC: https://azuredownloads.blob.core.windows.net/azuretoolkitforintellij/v3.20.0/azure-toolkit-for-intellij-2019.1.95.03-28-2019.zip

Steps:
1.	Right click on cluster “catest”, select “Delete”, the cluster is remoted from the list
2.	Refresh account “sparkwatchdog-c11”, the cluster “catest” show again.
3.	Delete “catest” for the second time

Result:
Get error with “null”

wezhang:
1. We can save last deleted cluster UUID to avoid refreshing show.
2. We can log or output more information from HTTP response for debugging.

t-rufang:
Looking.

t-rufang:
Fixed in #2967

v-jiche:
Verified. Fixed.
Build:  artifacts/azure-toolkit-for-intellij-2019.1.develop.1079.04-10-2019.zip

PR #2967, opened 2019-04-02
Fix deleted cluster re-appeared after refresh action issue

t-rufang:
- To address  #2958


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/common/azure/serverless/AzureSparkCosmosCluster.java
```diff
@@ -376,8 +376,23 @@ public String getName() {
...
     }
 
+    public boolean isStable() {
+        return isRunning()
+                && getMasterState() != null
+                && getMasterState().equalsIgnoreCase(SparkItemGroupState.STABLE.toString());
+    }
+
+    public boolean isRunning() {
+        return !getState().equalsIgnoreCase(SchedulerState.FINALIZING.toString())
```
wezhang: We can change the compare to enum type comparing later, rather than String.

PR #2994, opened 2019-04-10
Eclipse app service support remember user settings

zmssp:
App service in Eclipse support remember user settings.  We will fill the selection based on the user activities.
For example, when user open create dialog, we will fill the plan name, resource name, linux runtime based on the recently record. See the below pic.

image


PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.webapp/src/com/microsoft/azuretools/webapp/ui/AppServiceCreateDialog.java
```diff
@@ -1084,11 +1088,70 @@ protected void fillJavaVersion() {
         }
     }
 
+    private void fillUserSettings() {
+        try {
+            String os = CommonUtils.getPreference(CommonUtils.RUNTIME_OS);
+            if (os.equals(OperatingSystem.LINUX.toString())) {
```
jdneo: I'm thinking that using `equalIgnoreCase()` here would be better?

Issue #2996, opened 2019-04-10
The prompt message in Azure Sign out dialogue is wrong


sallyur:
### Environment:
```
Bits: Dev1077
OS: Windows10/Mac/Ubuntu
Eclipse: 2019-03
```

### Actual result:
It prompts "Signed in using file 'null"":
![image](https://user-images.githubusercontent.com/28860891/55856317-82438680-5b9c-11e9-84d5-be45b78936b6.png)

### Expect result:
It should display account message like IntelliJ:
![image](https://user-images.githubusercontent.com/28860891/55856425-db131f00-5b9c-11e9-8b27-a7f75ad9a70b.png)

bsaby:
The test passed.
Bits: daily724

Issue #2997, opened 2019-04-10
Device login failed with high probability


bsaby:
### Environment:
```
Bits: Dev1077
OS: Windows10/Mac/Ubuntu
Eclipse: 2019-03
```
### Repro Steps:
1. Open `Azure Sign In` dialog
2. Select `Device login` or `Service Principal`-> `New `login method
3. Click `Copy&Open` button in `Azure Device Login` dialog
4. Paste the code in the website
5. Type your account info

### Result:
The login is successful on the website but failed in Eclipse
![image](https://user-images.githubusercontent.com/22470312/55857009-97b9b000-5b9e-11e9-812d-1a4f7b662b89.png)
![image](https://user-images.githubusercontent.com/22470312/55857013-9ab4a080-5b9e-11e9-980b-5a45a7c231ee.png)

### Log
```
java.util.concurrent.ExecutionException: java.net.SocketTimeoutException: Read timed out
	at java.util.concurrent.FutureTask.report(Unknown Source)
	at java.util.concurrent.FutureTask.get(Unknown Source)
	at com.microsoft.azuretools.core.ui.DeviceLoginWindow$DeviceLoginDialog.pullAuthenticationResult(DeviceLoginWindow.java:181)
	at com.microsoft.azuretools.core.ui.DeviceLoginWindow$DeviceLoginDialog.lambda$0(DeviceLoginWindow.java:102)
	at java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.util.concurrent.FutureTask.run(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.lang.Thread.run(Unknown Source)
Caused by: java.net.SocketTimeoutException: Read timed out
	at java.net.SocketInputStream.socketRead0(Native Method)
	at java.net.SocketInputStream.socketRead(Unknown Source)
	at java.net.SocketInputStream.read(Unknown Source)
	at java.net.SocketInputStream.read(Unknown Source)
	at sun.security.ssl.InputRecord.readFully(Unknown Source)
	at sun.security.ssl.InputRecord.read(Unknown Source)
	at sun.security.ssl.SSLSocketImpl.readRecord(Unknown Source)
	at sun.security.ssl.SSLSocketImpl.readDataRecord(Unknown Source)
[...]

Issue #2995, opened 2019-04-10
Wrong font in Azure Device Login dialog


bsaby:
### Environment:
```
Bits: Dev1077
OS: Windows10/Mac/Ubuntu
Eclipse: 2019-03
```

### Result:
Windows:
![image](https://user-images.githubusercontent.com/22470312/55855730-e36a5a80-5b9a-11e9-8202-aab0f838af61.png)
Mac:
![image](https://user-images.githubusercontent.com/22470312/55855735-e5341e00-5b9a-11e9-8e14-bf102a05852d.png)

### Expect:
 We should use the same (UI) font everywhere.

wezhang:
As far as I know, it can't be since the different platform fonts have different copyrights. If we align it with IntelliJ UI font, it should be ok.

PR #3002, opened 2019-04-10
Fix issue #2996 and # 2995 and #2997

zmssp:
Fix issue #2996 and #2995 and #2997 
* Set the font family and size 
* Fix the logout prompt message
* Try 3 times when check the login status


PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.core/src/com/microsoft/azuretools/core/ui/DeviceLoginWindow.java
```diff
@@ -195,10 +201,10 @@ private void pullAuthenticationResult(final AuthenticationContext ctx, final Dev
 
         private String createHtmlFormatMessage() {
             final String verificationUrl = deviceCode.getVerificationUrl();
-            return "<body bgcolor=\"#F0F0F0\"><p>"
+            return "<div style=\"font-family: Arial; font-size: 13\"><body bgcolor=\"#F0F0F0\"><p>"
```
Flanker32: We may get the style info from code rather than hard code

PR #3026, opened 2019-04-15
Refine debug/run configuration from context menu.

konjac:
The change refines the behavior to select the run configuration for run/debug in the context menu.

-  If there is a selected config,  we will check 
1.  If the config type is current default spark type
2. If the config has the same main class with current context
If  1 and 2 are both true, this selected config will be used as run/debug config. 
- Otherwise, we will create a new config based on the current selected default spark type.


wezhang:
Closed for more discussion.

Issue #3024, opened 2019-04-15
[IntelliJ][2019.1][FailureDebug]Local Run and submit faild when use spark2.3.0 and spark2.3.2 to create failure debug project.


v-lanjli:
**Build:** https://vscjavaci.cloudapp.net/job/Azure_Toolkits_for_Java/job/AzureToolsManager_develop/1083/Azure/artifacts/azure-toolkit-for-intellij-2019.1.develop.1083.04-12-2019.zip 
**IntelliJ Version:** IntelliJ IDEA 2019.1 (Community Edition)
**OS:** Win\OS\Linux

**Steps:**

1. Open IntelliJ2019.1.
2. Create a new 'Spark Project with Failure Task Debugging Sample' project.
3. Select Spark 2.3.0 or Spark 2.3.2
4. Run or Submit '[Spark on Cosmos] AgeMean_Div0'

**Results:** Run and Submit were failded.
![image](https://user-images.githubusercontent.com/40199316/56122999-3468bd00-5fa6-11e9-966f-a5580f8f960b.png)

wezhang:
![image](https://user-images.githubusercontent.com/7385667/56175962-fada9500-602b-11e9-9543-62e26bd32ddd.png)
Packaged spark-tools v2.3.x libraries are bad.

v-lanjli:
Verified and Closed

PR #3029, opened 2019-04-16
Fix spark-tools v2.3.x packages classes missing issue

wezhang:
To address #3024


wezhang:
Confirmed in CI build packages:
image

PR #3046, opened 2019-04-17
Rules change when probe HDI new API

t-rufang:
For probe new HDI API, if the response:
 - Status code 403, we will take it as a HDI Reader cluster
 - Status code 200 with cluster credential provided, we take it as a HDI Owner cluster
 - Status code 200 without cluster credential provided, **telemetry will be sent**. Legacy API will be used. 
 - For network error, we will log the error, no telemtry will be sent. Legacy API will be used.
 - All the other errors will be shown as `HDInsightNewApiUnavailableException` and **telemetry will be sent**. Legacy API will be used.

### DIFF
--- a/Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/cluster/HDInsightNewAPI/ClusterOperationNewAPIImpl.java
+++ b/Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/cluster/HDInsightNewAPI/ClusterOperationNewAPIImpl.java
@@ -66,28 +66,45 @@ public class ClusterOperationNewAPIImpl extends ClusterOperationImpl implements 
             final SubscriptionDetail subscription,
             final String clusterId) throws IOException {
         return getClusterConfigurationRequest(subscription, clusterId)
-                .map(clusterConfiguration -> true)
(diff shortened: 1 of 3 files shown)


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/cluster/HDInsightNewAPI/ClusterOperationNewAPIImpl.java
```diff
@@ -66,28 +66,45 @@
...
+                        setRoleType(HDInsightUserRoleType.OWNER);
+                        return true;
+                    } else {
+                        final Map<String, String> properties = new HashMap<>();
+                        properties.put("ClusterID", clusterId);
+                        properties.put("StatusCode", "200");
+                        properties.put("ErrorDetails", "Cluster credential is incomplete.");
+                        AppInsightsClient.createByType(AppInsightsClient.EventType.Telemetry, this.getClass().getSimpleName(), null, properties);
+
+                        log().warn("Cluster credential is incomplete even if successfully get cluster credential.");
```
wezhang: Let's gather feedbacks from this error:

```suggestion
                        log().error("Cluster credential is incomplete even if successfully get cluster credential.");
```

PR #3065, opened 2019-04-19
Don't probe new API when the switch is off

t-rufang:


t-rufang:
Close this PR since it's already included in #3130

PR #3078, opened 2019-04-24
Expose storage root path property for linked HDInsight reader role cluster

t-rufang:


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/cluster/HDInsightAdditionalClusterDetail.java
```diff
@@ -46,6 +46,9 @@
     @Nullable
     private HDStorageAccount defaultStorageAccount;
 
+    @Nullable
+    private String defaultStorageRootPath;
```
wezhang: Is this a cluster property?

PR #3099, opened 2019-05-06
Use NetworkInterface get mac address when ifconfig/getmac fails

Flanker32:
Use NetworkInterface get mac address when `ifconfig`/`getmac` fails.

### DIFF
--- a/Utils/azuretools-core/src/com/microsoft/azuretools/azurecommons/util/GetHashMac.java
+++ b/Utils/azuretools-core/src/com/microsoft/azuretools/azurecommons/util/GetHashMac.java
@@ -89,19 +89,13 @@ public class GetHashMac {
             }
             ProcessBuilder probuilder = new ProcessBuilder(command);
             Process process = probuilder.start();
-            InputStream inputStream = process.getInputStream();
-            InputStreamReader inputStreamReader = new InputStreamReader(inputStream);
-            BufferedReader br = new BufferedReader(inputStreamReader);
-            String tmp;
-            while ((tmp = br.readLine()) != null) {
-                ret += tmp;
-            }
-
-            if (inputStream != null) {
-                inputStream.close();
-            }
-            if (br != null) {
-                br.close();
+            try (final InputStream inputStream = process.getInputStream();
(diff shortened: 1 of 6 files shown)


Utils/azuretools-core/src/com/microsoft/azuretools/azurecommons/util/GetHashMac.java
```diff
@@ -42,38 +53,40 @@ public static boolean IsValidHashMacFormat(String hashMac) {
...
     }
 
-    public static String GetHashMac() {
+    public static String getHashMac() {
         String ret = null;
-        String mac_raw = GetRawMac();
-        if (mac_raw != null && !mac_raw.isEmpty()) {
-            String mac_regex = "([0-9A-Fa-f]{2}[:-]){5}[0-9A-Fa-f]{2}";
+        String mac_raw = getRawMac();
+        mac_raw = isValidMac(mac_raw) ? mac_raw : getRawMacWithoutIfconfig();
```
wezhang: The return might by empty string. How to handle this?

Flanker32: Yes, we may lose the use if we can't get the mac with IP tools or Java NetworkInterface, but I think it's fine as there will be few users will meet this issue.

wezhang: How about to generate an UUID and put it into the `<user home>/AzureToolsForIntelliJ` folder?

Flanker32: @wezhang It should work to generate a UUID, but I am not sure how many users will meet this issue, they could send telemetries but we can't get the mac address. Besides, if we want to do this job, we may better to this in plugin initialization functions, for IntelliJ is [here](https://github.com/microsoft/azure-tools-for-java/blob/develop/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/AzurePlugin.java#L155). If get mac returns empty, generate a hash UUID installation id.

PR #3101, opened 2019-05-07
Improve app service data loading performance

zmssp:
When we get meta data from from the server side, most of the time we just wait for IO.
We should send this IO request concurrently.


Utils/azuretools-core/src/com/microsoft/azuretools/utils/AzureModelController.java
```diff
@@ -297,24 +304,57 @@ public static synchronized void updateResourceGroupMaps(IProgressIndicator progr
...
+                    }
+                    subscriber.onCompleted();
+                } catch (Exception e) {
+                    Exceptions.propagate(e);
+                } finally {
+                    countDownLatch.countDown();
+                }
+            }).subscribeOn(Schedulers.io()), sdSet.size()).subscribeOn(Schedulers.io()).subscribe();
+
+        new Thread(() -> {
```
andxu: Should test whether this thread will block process exit or not

zmssp: Set the thread to a daemon thread

PR #3109, opened 2019-05-08
Fix cannot detect default storage type for linked HDI reader cluster issue

t-rufang:


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/HDInsightHelper.java
```diff
@@ -36,7 +37,7 @@
     boolean isOptIn();
 
     @Nullable
-    default NodeActionListener createAddNewHDInsightReaderClusterAction(@NotNull HDInsightRootModule module, @NotNull String clusterName) {
+    default NodeActionListener createAddNewHDInsightReaderClusterAction(@NotNull HDInsightRootModule module, @NotNull ClusterDetail clusterDetail) {
```
wezhang: How about the eclipse side effect?

t-rufang: This action is designed for HDI reader cluster link action and is only called by IntelliJ. Therefore, no side effect for Eclipse.

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/serverexplore/action/AddNewHDInsightReaderClusterAction.kt
```diff
@@ -23,14 +23,33 @@
...
 
-class AddNewHDInsightReaderClusterAction(val module: HDInsightRootModule, val clusterName: String): NodeActionListener() {
+class AddNewHDInsightReaderClusterAction(val module: HDInsightRootModule, val selectedClusterDetail: ClusterDetail): NodeActionListener() {
     override fun actionPerformed(event: NodeActionEvent?) {
-        val form = AddNewHDInsightReaderClusterForm(module.project as Project, module, clusterName)
+        val defaultStorageRootPath = selectedClusterDetail.defaultStorageRootPath
+        val form = object: AddNewHDInsightReaderClusterForm(module.project as Project, module, selectedClusterDetail.name) {
+            override fun afterOkActionPerformed() {
+                val linkedCluster =
+                    ClusterManagerEx.getInstance().findClusterDetail({ clusterDetail ->
```
wezhang: Do we need to find it again?

PR #3147, opened 2019-05-28
Send telemetry when select path in VFS of ADLS Gen2

t-rufang:


lcadzy:
Thank you! Rui

lcadzy:
change name to "CHOOSE_REFERENCE_JAR_GEN2"

Issue #3154, opened 2019-05-29
[Eclipse]Launch config for the first time, reader role cluster does not have link cluster entry


jingyanjingyan:
Build: dev 1135

Repro Steps:
1. Sign in reader account
2. Right click on project, click submit to pops the config dialog

Result:
The reader role cluster does not have link cluster entry
After choose another reader role cluster, the link "link cluster" display well
![image](https://user-images.githubusercontent.com/36756024/58527851-f4346580-8206-11e9-9adb-619eebdf01e0.png)

t-rufang:
Fixed in #3160

jingyanjingyan:
Fix in dev 1137

PR #3160, opened 2019-05-29
Show errors if the first cluster in cluster lists is HDI reader cluster in job submission dialog in Eclipse

t-rufang:
To address #3154


lcadzy:
Rui,Thank you!

PR #3167, opened 2019-05-30
Fix linked HDI Reader cluster submit job fail issue

t-rufang:
To fix 2 issues:
1. #3165 If user submit to an HDI Reader role cluster(not linked), the error message is `User name may not be null`. What we expect is error message `No ambari permission to submit job to the selected cluster`
2. If user link the HDI Reader role cluster in submit job dialog and click submit button, submit fails still fails with `User name may not be null`. What we expect is job successfully run.

Issue #3165, opened 2019-05-30
[Eclipse][Reader Role] The warning message need to update when you submit a job to reader role cluster.


v-lanjli:
**Build:** https://vscjavaci.cloudapp.net/job/Azure_Toolkits_for_Java/job/AzureToolsManager_develop/1138/Azure/artifacts/WindowsAzurePlugin4EJ.develop.1138.05-30-2019.zip

**Repro Steps:**

Submit a job to reader role cluster (Confirm that the cluster is the reader role, and don't link it).
Check the warning message.

![image](https://user-images.githubusercontent.com/40199316/58615924-eb1ec380-82ee-11e9-816c-1a78afdf4122.png)

v-lanjli:
Verified and fixed in this build https://vscjavaci.cloudapp.net/job/Azure_Toolkits_for_Java/job/AzureToolsManager_develop/1140/Azure/artifacts/WindowsAzurePlugin4EJ.develop.1140.05-30-2019.zip
Closed this issue.

PR #3168, opened 2019-05-30
Fix linked HDI Reader cluster submit job fail issue

t-rufang:
Refer to PR #3167. To fix 2 bugs, and the second one is a severe bug.
1. #3165 If user submit to an HDI Reader role cluster(not linked), the error message is `User name may not be null`. What we expect is error message `No ambari permission to submit job to the selected cluster`
2. If user link the HDI Reader role cluster in submit job dialog and click submit button, submit fails still fails with `User name may not be null`. What we expect is job successfully run.


zmssp:
Do not merge

PR #3188, opened 2019-06-11
Fix link aris cluster port not editable

lcadzy:


lcadzy:
Since Aris cluster port is no longer fixed now , change the text filed to editable

PR #3206, opened 2019-07-01
Support IC-2019.2 EAP

lcadzy:

PR #3207, opened 2019-07-01
Support IDEA 192 EAP

wezhang:
IDEA 192EAP splits Java related APIs from IDEA production into
a standalone plugin - Java.

Refer to:
 - [IDEA-195719](https://youtrack.jetbrains.com/issue/IDEA-195719)
 - JetBrains/intellij-community@b9c5cd00681b0e763d0cf86ce1cffb55c1ca2ac5

It's required to add Java plugin into dependency modules.

There is also Scala plugin API changed related to SBT version getter.

## IMPORTANT ##

**This is a breaking change, the IDEA 191 and older versions builds will
be failed.**


wezhang:
Integrated into #3206

PR #3224, opened 2019-07-04
 Add telemetry for actions of nodes in Azure Explorer and serverless job view table actions

t-rufang:
### Code Refactor:
1. Remove unused field `name` for class `NodeActionListener` since it's never used.  Remove constructor `public NodeActionListener(Node node) {}` for class `NodeActionListener` since nothing is done in this constructor. We can call `public NodeActionListener() {}` instead.
2. Notice that every action in one `Node` shares the same operation name, we can implement a `getServiceName` method in class `Node`, and get service name just like the following way in `NodeActionListener`. Therefore, as you can see, we add a parameter `NodeActionEvent event` for `NodeActionListener::getServiceName`.
```
    protected String getServiceName(NodeActionEvent event) {
        try {
            return event.getAction().getNode().getServiceName();
        } catch (Exception ignore) {
            return TelemetryConstants.ACTION;
        }
    }
```
3. Slight change of `NodeActionListener::getOperationName` to align to the format of operation name in `TelemetryConstants`.
```
    protected String getOperationName(NodeActionEvent event) {
        try {
            return event.getAction().getName().toLowerCase().trim().replace(" ", "-");
        } catch (Exception ignore) {
            return "";
        }
    }
``` 
4. Remove all the `addClickActionListener` call in sub-class of `Node` class. Instead, we put those action into `onNodeClick` method. As you can see, `clickAction` is already bind to an action listener in class `Node`.

```
    protected void loadActions() {
        // add the click action handler
        addClickActionListener(new NodeActionListener() {
            @Override
            public void actionPerformed(NodeActionEvent e) {
                onNodeClick(e);
            }
        });
        // Other not related code are ignored...
}

    public void addClickActionListener(NodeActionListener actionListener) {
        clickAction.addListener(actionListener);
    }
```
If we bind another listener to `clickAction`, duplicate telemetries will be sent when each listener calls `actionPerformedAsync`. Check the definition for `NodeAction::fireNodeActionEvent`
```
    public void fireNodeActionEvent() {
        if (!listeners.isEmpty()) {
            final NodeActionEvent event = new NodeActionEvent(this);
            for (final NodeActionListener listener : listeners) {
                listener.beforeActionPerformed(event);
                Futures.addCallback(listener.actionPerformedAsync(event), new FutureCallback<Void>() {
                    @Override
                    public void onSuccess(Void aVoid) {
                        listener.afterActionPerformed(event);
                    }

                    @Override
                    public void onFailure(Throwable throwable) {
                        listener.afterActionPerformed(event);
                    }
                });
            }
        }
    }

```

### DIFF
--- a/Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/docker/DockerContainerNode.java
+++ b/Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/docker/DockerContainerNode.java
@@ -256,7 +256,7 @@ public class DockerContainerNode extends AzureRefreshableNode implements Telemet
     }
 
     @Override
-    protected String getServiceName() {
+    protected String getServiceName(NodeActionEvent event) {
       return DOCKER;
     }
 
--- a/PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.azureexplorer/src/com/microsoft/azuretools/azureexplorer/actions/CreateArmVMAction.java
+++ b/PluginsAndFeatures/azure-toolkit-for-eclipse/com.microsoft.azuretools.azureexplorer/src/com/microsoft/azuretools/azureexplorer/actions/CreateArmVMAction.java
@@ -33,7 +33,7 @@ import com.microsoft.tooling.msservices.serviceexplorer.azure.vmarm.VMArmModule;
 @Name("Create VM")
 public class CreateArmVMAction extends NodeActionListener {
     public CreateArmVMAction(VMArmModule node) {
-        super(node);
+        super();
     }
 
     @Override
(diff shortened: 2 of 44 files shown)


Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/docker/DockerContainerNode.java
```diff
@@ -256,7 +256,7 @@ protected void onSubscriptionsChanged(NodeActionEvent e) throws AzureCmdExceptio
     }
 
     @Override
-    protected String getServiceName() {
+    protected String getServiceName(NodeActionEvent event) {
       return DOCKER;
```
Flanker32: Some of Node uses static import while others specifying the class, maybe we could unify the code style later.

Issue #3197, opened 2019-06-18
[IntelliJ][ReportedByUser] HDInsight Explorer


superPusher:
IntelliJ build version: 2018.1.8 IC-181.5684.4
OS: Windows 7
JDK: JetBrains s.r.o 1.8.0_152-release
Plugin version: 3.13.0-2018.1
Error message: 
```
com.microsoft.azure.hdinsight.sdk.common.HDIException: Failed to get default storage account name
	at com.microsoft.azure.hdinsight.sdk.cluster.ClusterDetail.getDefaultStorageAccount(ClusterDetail.java:261)
	at com.microsoft.azure.hdinsight.sdk.cluster.ClusterDetail.getConfigurationInfo(ClusterDetail.java:227)
	at com.microsoft.azure.hdinsight.serverexplore.hdinsightnode.StorageAccountFolderNode.refreshItems(StorageAccountFolderNode.java:55)
	at com.microsoft.tooling.msservices.serviceexplorer.RefreshableNode.refreshItems(RefreshableNode.java:96)
	at com.microsoft.tooling.msservices.serviceexplorer.RefreshableNode$2.run(RefreshableNode.java:148)
	at com.microsoft.intellij.helpers.IDEHelperImpl$2$1.run(IDEHelperImpl.java:157)
	at com.intellij.openapi.progress.impl.CoreProgressManager$TaskRunnable.run(CoreProgressManager.java:750)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$runProcess$1(CoreProgressManager.java:157)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:580)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:525)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.executeProcessUnderProgress(ProgressManagerImpl.java:85)
	at com.intellij.openapi.progress.impl.CoreProgressManager.runProcess(CoreProgressManager.java:144)
	at com.intellij.openapi.progress.impl.CoreProgressManager$4.run(CoreProgressManager.java:395)
	at com.intellij.openapi.application.impl.ApplicationImpl$1.run(Applic

wezhang:
It would be better to record more cluster details, such as type, version

lcadzy:
#3235

PR #3235, opened 2019-07-08
Wrap more details in HDIExp when getting storage account name

lcadzy:


lcadzy:
#3197

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/cluster/ClusterDetail.java
```diff
@@ -370,56 +371,60 @@ private IHDIStorageAccount getDefaultStorageAccount(Map<String, String> coresite
...
+                    String fileSystem = pathInfo.path.getUserInfo();
+                    return new ADLSGen2StorageAccount(this, accountName, null, true, fileSystem, pathInfo.path.getScheme());
 
-            default:
-                return null;
+                default:
+                    return null;
+            }
+        } catch (HDIException ex) {
+            throw new HDIException(String.format("Encounter exception when getting storage configuration for root path %s with type %s",
```
wezhang: Can it be wrapped as a classified Exception?

lcadzy: No , all the methods call this function will catch the exception and return empty or null and this exception can't reach to the classified exception stage.

PR #3243, opened 2019-07-09
Enable syntax highlight for Spark Cosmos Serverless job detail JSON

wezhang:
JSON Syntax highlight result:
image


wezhang:
Addressed #2632

PR #3258, opened 2019-07-10
[on hold] Move customer survey to qualtrics.

Flanker32:
Move customer survey from github issue to [qualtrics](https://microsoft.qualtrics.com/jfe/form/SV_b17fG5QQlMhs2up).


yucwan:
@wezhang @Flanker32 Qualtrics survey link should not be added here. We are collecting info like os or plugin version using the url parameters just like what we are doing with github issues. Instead, i think we should put the following links here, as i already added for our docs:

 To report bugs or request new features, create issues on [our GitHub repository](https://github.com/Microsoft/azure-tools-for-java/issues). Or, ask questions on [Stack Overflow with tag azure-java-tools](https://stackoverflow.com/questions/tagged/azure-java-tools)。

PR #3259, opened 2019-07-10
Add telematry for run config file

t-rufang:
Design of telemetry for java swing components:
If you want to add telemetry for some components in a panel. you need to follow these steps:
1. `setName()` for those components that you want to send telemetry for in the panel. The name will become part of the telemetry properties.
```
    private val enableRemoteDebugCheckBox = JCheckBox("Enable Spark remote debug", true).apply {
        name = "enableRemoteDebugCheckBox"
    }
```
2. Specify `serviceName`(like "HDInsight") and add telemetry listener for the whole panel.
```
        myPanel.addTelemetryListener("HDInsight")
```
3. We have implement telemetry listeners for some type of components that we have interest in. You can implement your own listener in `ContainerTelemetryExtension.kt`.
WARNING: If class A is sub-class of class B, we should put matcher to class A in front of matcher to class B.
```
            is JCheckBox -> component.addActionListener {
                val props = mapOf("isSelected" to component.isSelected.toString())
                createLogEvent(serviceName, "check-check-box", component.name, props)
            }
```

In this example, when the check box is checked, we will send a telemetry with the following properties:
 - serviceName: "HDInsight"
 - operationName: "check-check-box"
 - componentName: "enableRemoteDebugCheckBox"
 - isSelected: "true"
 - operationTime: "2019-07-09T02:47:34.245Z"

Please check class `ContainerTelemetryExtensionTest` for a complete sample.


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/telemetry/IdeaTelemetryUtils.kt
```diff
@@ -0,0 +1,143 @@
...
+import com.microsoft.azuretools.telemetrywrapper.EventUtil
+import com.microsoft.intellij.ui.components.JsonEnvPropertiesField
+import java.awt.Component
+import java.awt.event.*
+import java.time.Instant
+import java.time.format.DateTimeFormatter
+import javax.swing.*
+
+object IdeaTelemetryUtils {
+    fun addComponentListener(component: Component, serviceName: String) {
```
wezhang: I would suggest Kotlin extended function for it.

wezhang: And the name can be `addTelemetryListener`

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/spark/common/SparkSubmitResponse.java
```diff
@@ -60,4 +62,14 @@ public boolean isAlive() {
                 !this.getState().equals("success") &&
                 !this.getState().equals("dead");
     }
+
+    @Nullable
+    static public SparkSubmitResponse parseJSON(String json) {
```
wezhang: Who calls it?

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/forms/dsl/FormBuilder.kt
```diff
@@ -87,10 +88,13 @@ class FormBuilder(var columnTemplate: ColumnTemplate? = null) {
 
     fun createGridLayoutManager() = GridLayoutManager(rowSize, colSize)
 
-    fun buildPanel(): JPanel = JPanel().apply {
+    fun buildPanel(serviceName: String): JPanel = JPanel().apply {
```
wezhang: May not be a good solution to add into a tool class. Let's consider do it in component extend functions.

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/spark/common/SparkSubmitResponse.java
wezhang: We can move it to test classes if is only called by them.

PR #3396, opened 2019-08-13
Fix can't select the correct cluster as expected bug

t-rufang:
- For Arcadia Spark Compute, cluster displays as "computeName@workspaceName", but when we `findClusterbyId`, we only try to match cluster name from candidate clusters, which is not accurate.


lcadzy:
Thank you rui!

PR #3397, opened 2019-08-13
Place null in catch, because it will return null when an exception ha…

pi1024e:
…ppens anyway.


vscjavaci:

pi1024e:
You posted this the second I clicked "Send Pull Request". Are you a bot or person?

Flanker32:
@t-rufang Please have a look.

t-rufang:
@pi1024e This PR doesn't change any behaviour of the method `convertToObjectQuietly`. If an IOException happens in this method, this method will swallow the exception and return `null` no matter with or without this patch. I wonder you want to add this patch? Have you ever run into any exception that you want to catch?

t-rufang:
Hi @pi1024e , I will close this issue since there are no updates. Feel free to reopen it if you have other concerns.

PR #3430, opened 2019-08-19
Fix NPE issue in auth

Flanker32:
- Fix toolwindow listener will not close when close projects
- Fix annoying prompt when account has no subscription


Flanker32:
Close for git branch issue

PR #3456, opened 2019-08-22
Move Comparable class out as a seperate interface for IClusterDetail

t-rufang:
- This is a `Code Refactor` PR


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/cluster/ComparableCluster.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.microsoft.azuretools.azurecommons.helpers.NotNull;
+
+public interface ComparableCluster extends Comparable<ComparableCluster> {
+    @NotNull
+    String getClusterIdForConfiguration();
+
+    @Override
+    default int compareTo(@NotNull ComparableCluster other) {
+        return this.getClusterIdForConfiguration().compareToIgnoreCase(other.getClusterIdForConfiguration());
```
wezhang: We can have this as a shortcut:

```
        if (this == other) {
            return 0;
        }
```

PR #3472, opened 2019-08-27
Fix survey window may appear multi times when user open multi intellij window

Flanker32:
Fix survey window may appear multi times when user open multi intellij window


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/helpers/CustomerSurveyHelper.java
```diff
@@ -100,7 +101,11 @@ public void neverShowAgain() {
         sendTelemetry(TELEMETRY_VALUE_NEVER_SHOW);
     }
 
-    private boolean isAbleToPopUpSurvey() {
+    private synchronized boolean isAbleToPopUpSurvey() {
+        if (isShown) {
+            return false;
+        }
+        isShown = true;
```
andxu: when to set isShown  to false? never?

Flanker32:
``` java
private boolean isShown = false;
```
And it will set to false once the window is pop up and will never show again unless users restart the IDE

PR #3526, opened 2019-09-11
Remove focus of editorPanel in case focus loss

Flanker32:
Remove focus of editorPanel in case focus loss, for issue #3350


vscjavaci:

Issue #3513, opened 2019-09-09
[intelliJ][spark interactive console]Get error when run livy interactive console for airs cluster.


v-xiaocw:
Build:
azure-toolkit-for-intellij-2019.2.develop.1234.09-05-2019

Repro Steps:
1.Create a SQL Server Big Data Cluster configuration, choose spark cluster and main class.
2.Run livy interactive console from Tools/spark console.
Result:
Get error message.
![image](https://user-images.githubusercontent.com/43333874/64527478-e937a980-d2ba-11e9-872d-713bac07693b.png)

t-rufang:
Fixed in #3529

Issue #3510, opened 2019-09-09
[intelliJ][spark interactive console]Get error when run livy interactive console for hdinsight and cosomos job.


v-jiche:
Build:
azure-toolkit-for-intellij-2019.2.develop.1234.09-05-2019 (4)

Repro Steps:
1. Create a HDInsight configuration, choose spark cluster and main class.
2. Run livy interactive console from Tools/spark console.
Result:
Get error message.
![image](https://user-images.githubusercontent.com/35950097/64512031-19f9fc00-d318-11e9-83df-e22d73d4bc59.png)
3. Run livy interactive console from Tools/spark console for cosmos job.
Result:
Get error message.
![image](https://user-images.githubusercontent.com/35950097/64517960-e02ef280-d323-11e9-97d1-35ee6cc5e993.png)

v-jiche:
Bug repros in build: develop 1247
1. run interactive console for Cosmos job:
Frequency: sometimes
![image](https://user-images.githubusercontent.com/35950097/64759279-46588700-d569-11e9-8038-899a3256dafe.png)

wezhang:
@v-jiche , I suppose it's been fixed, could you help to verify it? Thanks!

v-jiche:
Verified. Fixed.
Build: azure-toolkit-for-intellij-2019.2.4.develop.1342.01-08-2020.zip

PR #3529, opened 2019-09-11
Fix fail to launch Livy interactive console for HDInsight and Aris cluster issue

t-rufang:
To address #3510 and #3513


vscjavaci:

t-rufang:
> Can one of the admins verify this patch?

Looking.

Issue #3372, opened 2019-08-01
[IntelliJ][ReportedByUser] Uncaught Exception  java.lang.NullPointerException


v-xiaocw:
IntelliJ build version: 2019.2 IC-192.5728.98
OS: Windows 10
JDK: JetBrains s.r.o 11.0.3
Plugin version: 3.25.0-2019.2
Additional Info: None
Build: dev 1196
Repro Steps:
1.Link a cluster
2.Expand the cluster drop down menu
3.Click the Storage Accounts
Result:
Pops up the  red exclamation mark
Show the following:
Parent component: 
```
com.intellij.openapi.wm.impl.IdeRootPane[,8,0,1384x771,invalid,layout=com.intellij.openapi.wm.impl.IdeRootPane$MyRootLayout,alignmentX=0.0,alignmentY=0.0,border=,flags=449,maximumSize=,minimumSize=,preferredSize=]
```
Error message: 
java.lang.NullPointerException
	at java.desktop/javax.swing.ImageIcon.<init>(ImageIcon.java:217)
	at com.microsoft.azuretools.ijidea.ui.WarningMessageForm.$$$setupUI$$$(WarningMessageForm.java)
	at com.microsoft.azuretools.ijidea.ui.WarningMessageForm.<init>(WarningMessageForm.java:48)
	at com.microsoft.azure.hdinsight.common.HDInsightHelperImpl$6$1.<init>(HDInsightHelperImpl.java:254)
	at com.microsoft.azure.hdinsight.common.HDInsightHelperImpl$6.run(HDInsightHelperImpl.java:254)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.doRun(LaterInvocator.java:433)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.runNextEvent(LaterInvocator.java:416)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.run(LaterInvocator.java:399)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:313)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:776)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:727)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:721)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
[...]

PR #3546, opened 2019-09-17
Fix cannot find warning icon caused NPE issue

t-rufang:
- The RCA is that our code is trying to use an icon from JDK8. However, the default IntelliJ 2019.2 is built on top of JDK11, which doesn't have the icon and causes the NPE bug. If we [run the IntelliJ action](https://www.jetbrains.com/help/idea/searching-everywhere.html#search_actions) `Switch boot jdk`, switch JDK to JDK8 and restart IntelliJ, we will find the error missing.
 - To fix this bug, instead we use warning icon from IntelliJ platform SDK rather than JDK.
 - To address #3391, #3392, #3372, #3814


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azuretools/ijidea/ui/WarningMessageForm.java
```diff
@@ -56,6 +57,10 @@ public WarningMessageForm(
         }
```
wezhang: Let's avoid to call under initialization methods from constructor.

PR #3603, opened 2019-09-24
Update Azure client runtime and authentication library to 1.6.13

wezhang:


Flanker32:
Shall we update the version in Utils as well?

wezhang:
> Shall we update the version in [Utils ](https://github.com/microsoft/azure-tools-for-java/blob/ce66517fefcd30c6bd6f60e222f8b72c5f94e8d6/Utils/pom.xml#L342)as well?

np, will come with following patches.

Issue #3639, opened 2019-09-29
[IntelliJ][ReportedByUser] NPE at com.microsoft.azure.hdinsight.spark.ui.SparkSubmissionJobUploadStorageWithUploadPathPanel.setData


smithyhammer:
IntelliJ build version: 2019.2.3 IU-192.6817.14
OS: Windows 10
JDK: JetBrains s.r.o 11.0.4
Plugin version: 3.27.0-2019.2
Additional Info: None
Parent component: 
```
com.intellij.openapi.wm.impl.IdeRootPane[,8,8,1921x1079,layout=com.intellij.openapi.wm.impl.IdeRootPane$MyRootLayout,alignmentX=0.0,alignmentY=0.0,border=,flags=449,maximumSize=,minimumSize=,preferredSize=]
```
Error message: 
```
IdeaLoggingEvent[message=null, throwable=java.lang.IllegalStateException: Exception thrown on Scheduler.Worker thread. Add `onError` handling.

Caused by: rx.exceptions.OnErrorNotImplementedException
Caused by: java.lang.NullPointerException
	at com.microsoft.azure.hdinsight.spark.ui.SparkSubmissionJobUploadStorageWithUploadPathPanel.setData(SparkSubmissionJobUploadStorageWithUploadPathPanel.kt:373)
	at com.microsoft.azure.hdinsight.spark.ui.SparkSubmissionJobUploadStorageWithUploadPathPanel$ViewModel$validateStorageInfo$6.call(SparkSubmissionJobUploadStorageWithUploadPathPanel.kt:329)
	at com.microsoft.azure.hdinsight.spark.ui.SparkSubmissionJobUploadStorageWithUploadPathPanel$ViewModel$validateStorageInfo$6.call(SparkSubmissionJobUploadStorageWithUploadPathPanel.kt:115)
Caused by: rx.exceptions.OnErrorThrowable$OnNextValue: OnError while emitting onNext value: com.microsoft.azur

fengshuaihu:
Please update to the latest version and try again, if you encounter this issue again, feel free to turn it on again

PR #3648, opened 2019-10-09
Enable setting null value to java swing component property

t-rufang:
To addess #3639


t-rufang:
Since the issue is not reproed in local environment, we will close this PR to avoid other unexpected errors.

PR #3670, opened 2019-10-21
Fix the Spark console uncaught exception for session service is done

wezhang:


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/spark/console/SparkLivySessionOutputStream.kt
```diff
@@ -41,13 +41,15 @@ class SparkLivySessionOutputStream(val session: Session) : ByteArrayOutputStream
 
         session.runCodes(codes)
                 .doOnEach { reset() }
+                .toBlocking()
```
t-rufang: Will it block UI thread if we make this thread blocking

wezhang: Yes, it will block the current thread a little ... And the impact is bigger than I imaged. Thanks!

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/spark/console/SparkConsoleExecuteAction.kt
```diff
@@ -85,14 +89,29 @@ class SparkConsoleExecuteAction() : AzureAnAction(), DumbAware, ILogger {
 
         // Send to process as a whole codes block
         val normalizedCodes = text.trimEnd() + "\n"
-        try {
+        val codeHint = normalizedCodes.split("\n").first()
```
t-rufang: The code might be only one line but with a long string. In this case the process bar will be messy. Maybe you can test this case first to see if it works well. If not, suggest to truncate the string to limited length.

wezhang: So far so good:
spark_console_too_long_progress

PR #3764, opened 2019-11-19
Support log aggregation for job with esp cluster

lcadzy:


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/spark/common/SparkBatchEspMfaSubmission.java
```diff
@@ -33,6 +34,7 @@
 
 public class SparkBatchEspMfaSubmission extends SparkBatchSubmission {
     public static final String resource = "https://hib.azurehdinsight.net";
+    private static final String authScheme = "Bearer ";
```
wezhang: No need for this, just inline should be ok.

Issue #3693, opened 2019-10-30
[IntelliJ][2019.3 EAP][console]Results is wrong when use command 'spark.sessionState' to run interactive console with link a cluster


v-lanjli:
**IntelliJ build version:** 2019.3 EAP IC-193.4778.7
**OS:** Mac/Win/Linux
**JDK:** JetBrains s.r.o 11.0.4
**Plugin version:** 3.28.0-LATEST-EAP-SNAPSHOT

**Repro Steps:**

1. Open IntelliJ and install the EAP plugin
2. Link a cluster 
3. Create config, and select a linked cluster
4. Tools -> Spark Console -> Run Spark Livy Interactive Session Console(Scala)
5. Enter 'spark.sessionState', press Ctrl+Enter to run
6. Check the result

**Actual**: Show ‘res3: String = Unnamed >> Spark Livy Interactive Session Console(Scala)spark.sessionState’
![image](https://user-images.githubusercontent.com/40199316/67839153-f2f8c300-fb2d-11e9-9957-8e2590da3f83.png)
**Expectation:** Spark> spark.sessionState
res3: org.apache.spark.sql.internal.SessionState = org.apache.spark.sql.internal.SessionState@2723a65d

v-lanjli:
Verified and fixed, closed this issue.
![image](https://user-images.githubusercontent.com/40199316/70131212-de739180-16bc-11ea-8cc5-1c017d84ed4f.png)

Issue #3686, opened 2019-10-30
[IntelliJ][2019.3 EAP][console]Layout issue with the results of the interactive console.


v-lanjli:
**IntelliJ build version:** 2019.3 EAP IC-193.4778.7
**OS:** Mac/Win/Linux
**JDK:** JetBrains s.r.o 11.0.4
**Plugin version:** 3.28.0-LATEST-EAP-SNAPSHOT

**Repro Steps:**

1. Open IntelliJ and install the EAP plugin
2. Link a cluster or sign in to Azure
3. Create config, and select a linked cluster for Spark Cluster 
4. Tools -> Spark Console -> Run Spark Livy Interactive Session Console(Scala)
5. Enter 'sc.appName' or 'spark.sessionState'
6. Check the result

**Actual:** 
1. Run interactive console with **link a cluster**: The command is displayed after the result 
![image](https://user-images.githubusercontent.com/40199316/67834553-a5c32400-fb22-11e9-8b5d-da4ad4c8e927.png)
2. Run interactive console by **sign in to Azure and select a cluster**: Run command is not highlighted.
![image](https://user-images.githubusercontent.com/40199316/67834940-ab6d3980-fb23-11e9-8b16-0c35bc503b66.png)

Expectation: The layout of the results should be consistent with below:
![image](https://user-images.githubusercontent.com/40199316/67835131-32baad00-fb24-11e9-8dcc-91db89014dee.png)

v-lanjli:
@wezhang The layout issue has been fixed, but there are still some issues:
1. In 192, the wording is **spark**, but in 193, the wording is **scala**.
![image](https://user-images.githubusercontent.com/40199316/70130738-04e4fd00-16bc-11ea-83f3-a960bf050b38.png)
2. The **scala** is not highlight.

v-lanjli:
Verified and fixed, closed this issue.

PR #3783, opened 2019-11-26
Fix Spark Livy console 193EAP regression

wezhang:
To address #3686, #3693


wezhang:
Committed as hot fix.

PR #3787, opened 2019-11-26
Fix auth card not sync with cluster type change

lcadzy:


lcadzy:
#3778

lcadzy:
Good job Yi!
Hope you can have a great journey.

PR #3792, opened 2019-12-02
add batch parameter to mvn cli

andxu:


Eskibear:
LGTM, but why you need a "-B" for non-interactive goals like `install` and `clean` ?

PR #3800, opened 2019-12-03
Be able to provide bundle bits

wezhang:


PluginsAndFeatures/azure-toolkit-for-intellij/gradle.properties
```diff
@@ -1,7 +1,7 @@
 javaVersion=1.8
 org.gradle.jvmargs='-Duser.language=en'
 sources=false
-intellij_version=IC-LATEST-EAP-SNAPSHOT
+intellij_version=IC-2019.2.4
```
t-rufang: quick question, why we keep the `intellij_version` as the latest EAP version before?

wezhang: Because we keep two version build before.

PR #3839, opened 2019-12-11
Deprecate storage explorer for HDInsight in both IntelliJ and Eclipse

t-rufang:


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/serverexplore/hdinsightnode/StorageAccountNode.java
```diff
@@ -23,109 +23,69 @@
...
-            log().warn(errorHint + ExceptionUtils.getStackTrace(ex));
-
-            DefaultLoader.getUIHelper().showError(errorHint + exceptionMsg, "HDInsight Explorer");
+    protected void loadAdditionalActions() {
+        if (clusterDetail instanceof ClusterDetail) {
+            addAction("Open Storage in Azure Management Portal", new NodeActionListener() {
+                @Override
+                protected void actionPerformed(NodeActionEvent e) throws AzureCmdException {
+                    try {
+                        final AzureManager azureManager = AuthMethodManager.getInstance().getAzureManager();
```
Flanker32: Shall we use openResourcesInPortal(HasId , String)` in `Node` here? So that node could share codes to open resources in portal

t-rufang: Resolved in the new iteration.

PR #3911, opened 2020-01-09
Abstract Azure Storage URI with AdlUri added

wezhang:
`getRoot()` method is removed with introduce several resolve methods

### DIFF
--- a/Utils/hdinsight-node-common/Test/java/com/microsoft/azure/hdinsight/common/AbfsUriScenario.kt
+++ b/Utils/hdinsight-node-common/Test/java/com/microsoft/azure/hdinsight/common/AbfsUriScenario.kt
@@ -75,30 +75,13 @@ class AbfsUriScenario {
         }
     }
 
-    @Then("^the Gen two base restful path should be$")
-    fun gen2BaseRestfulPathShouldBe(expectedBaseRestfulPath: DataTable) {
-        val actualBaseRestfulPath = this.restfulGen2Paths.stream()
-            .map { path ->
-                try {
-                    AbfsUri.parse(path).root.url.toString()
-                } catch (ex: UnknownFormatConversionException) {
-                    "invalid Gen2 URI"
-                }
-            }
-            .collect(Collectors.toList())
-
-        (0 until actualBaseRestfulPath.size).forEach {
-            assert(actualBaseRestfulPath[it] == expectedBaseRestfulPath.asList(String::class.java)[it])
-        }
-    }
-
     @Then("^the Gen two directory param should be$")
     fun gen2SubPathShouldBe(expectedSubPath: DataTable) {
         val actualSubPath = this.abfsUris.stream()
             .map { path ->
                 try {
                     AbfsUri.parse(path).directoryParam.toString()
-                } catch (ex: UnknownFormatConversionException) {
+                } catch (ex: Throwable) {
                     "invalid Gen2 URI"
                 }
             }
--- a/Utils/hdinsight-node-common/Test/java/com/microsoft/azure/hdinsight/common/AdlUriScenario.kt
+++ b/Utils/hdinsight-node-common/Test/java/com/microsoft/azure/hdinsight/common/AdlUriScenario.kt
@@ -32,6 +32,7 @@ class AdlUriScenario {
     data class AdlUriParameterEntry(val uri: String, val path: String, val storageName: String)
     data class UrlUriEqualEntry(val src: String, val dest: String, val isEqualed: Boolean)
     data class UriPathResolveAsRootEntry(val uri: String, val path: String, val result: String)
+    data class UriRelativizeCheckEntry(val src: String, val dest: String, val result: String)
 
     @Then("convert Gen ONE URL restful path to URI should be")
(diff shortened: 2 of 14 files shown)


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/AzureStorageUri.java
```diff
@@ -0,0 +1,124 @@
...
+     * Getter of raw URI
+     *
+     * @return the raw URI
+     */
+    public URI getRawUri() {
+        return rawUri;
+    }
+
+    /**
+     * Get URI with specified scheme, not limited to HTTP or HTTPS
```
t-rufang: When you say `not limited to HTTP or HTTPS`, do you mean `getUri()` can return URIs not only starting with `HTTP` or `HTTPS` but also other schemes?

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/WasbUri.java
```diff
@@ -43,29 +44,26 @@
...
-                getContainer(), getStorageAccount(), getEndpointSuffix(), getRelativePath()));
+                getContainer(), getStorageAccount(), getEndpointSuffix(), getPath()));
     }
 
+    @Override
     public URL getUrl() {
         try {
             return URI.create(String.format("https://%s.blob.%s/%s%s",
-                    getStorageAccount(), getContainer(), getEndpointSuffix(), getRelativePath())).toURL();
+                    getStorageAccount(), getContainer(), getEndpointSuffix(), getPath())).toURL();
```
t-rufang: I suppose the order should be `getStorageAccount(), getEndpointSuffix(), getContainer(), getPath()`. Could you also add UT for class `WasbUri`

Utils/hdinsight-node-common/Test/java/com/microsoft/azure/hdinsight/common/AbfsUriScenario.kt
```diff
@@ -75,30 +75,13 @@ class AbfsUriScenario {
...
-    }
-
     @Then("^the Gen two directory param should be$")
     fun gen2SubPathShouldBe(expectedSubPath: DataTable) {
         val actualSubPath = this.abfsUris.stream()
             .map { path ->
                 try {
                     AbfsUri.parse(path).directoryParam.toString()
-                } catch (ex: UnknownFormatConversionException) {
+                } catch (ex: Throwable) {
```
t-rufang: Why do we change it to `Throwable` here?

Utils/hdinsight-node-common/Test/java/com/microsoft/azure/hdinsight/common/AdlUriScenario.kt
```diff
@@ -0,0 +1,100 @@
...
+import cucumber.api.java.en.Then
+import java.util.*
+import kotlin.test.assertEquals
+import kotlin.test.assertNotEquals
+
+class AdlUriScenario {
+    data class UrlUriEntry(val url: String, val uri: String)
+    data class AdlUriParameterEntry(val uri: String, val path: String, val storageName: String)
+    data class UrlUriEqualEntry(val src: String, val dest: String, val isEqualed: Boolean)
+    data class UriPathResolveAsRootEntry(val uri: String, val path: String, val result: String)
```
t-rufang: Using data class makes both `.feature` file and test code much simpler!

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/AzureStorageUri.java
wezhang: Yes, such as wasbs://, adl://, abfs:// and so on.

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/WasbUri.java
wezhang: Good catch!

Utils/hdinsight-node-common/Test/java/com/microsoft/azure/hdinsight/common/AbfsUriScenario.kt
wezhang: There is another Throwable here: `IllegalArgumentException`, and any throwables are not allowed here.

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/AzureStorageUri.java
t-rufang: I suppose `getUri` should return URIs not starting with `http` because we need to differenciate it with `getUrl`.

Utils/hdinsight-node-common/Test/java/com/microsoft/azure/hdinsight/common/AbfsUriScenario.kt
t-rufang: I see. I will change all the `UnknownFormatConversionException` to `Throwable` in my PR.

PR #3922, opened 2020-01-14
Fix storage type switch issues when selecting a different cluster and other UX issues

wezhang:


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/cluster/ClusterDetail.java
```diff
@@ -391,20 +390,17 @@ private IHDIStorageAccount getDefaultStorageAccount(Map<String, String> coresite
...
-                            defaultStorageRootPath, pathInfo.storageType));
-                }
-
-                String defaultContainerName = pathInfo.path.getUserInfo();
-                String defaultStorageAccountKey = StorageAccountKeyPrefix + storageAccountName;
+                WasbUri wasbUri = WasbUri.parse(defaultStorageRootPath);
+                String storageAccountName = wasbUri.getStorageAccount();
+                String defaultContainerName = wasbUri.getContainer();
+                String defaultStorageAccountKey =
+                        StorageAccountKeyPrefix + storageAccountName + ".blob." + wasbUri.getEndpointSuffix();
```
t-rufang: Maybe we can change the regex pattern in `WasbUri` to include `.blob.` in the endpointSuffix field, therefore here the code could be `StorageAccountKeyPrefix + storageAccountName + wasbUri.getEndpointSuffix()`

Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/sdk/storage/StorageAccountType.java
```diff
@@ -21,9 +21,41 @@
...
+    public static StorageAccountType parseUri(String uri) {
+        try {
+            return parseUri(URI.create(uri));
+        } catch (IllegalArgumentException ignored) {
+            return UNKNOWN;
+        }
+    }
+
+    public static StorageAccountType parseUri(URI uri) {
+        if (startsWithAny(uri.getScheme().toLowerCase(), "http", "https")) {
```
t-rufang: Currently all of the AzureStorageUri can handle `http` and `https` pattern, maybe we can remove this check here

t-rufang:
Close this PR since the code in this PR is seperated into another two PRs #3947 and #3960

PR #3935, opened 2020-01-15
Clean up HDInsight clusters from Azure explorer after signed out

wezhang:


Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/ClusterManagerEx.java
```diff
@@ -77,6 +77,11 @@ public static ClusterManagerEx getInstance() {
             synchronized (ClusterManagerEx.class) {
                 if (instance == null) {
                     instance = new ClusterManagerEx();
+
+                    AuthMethodManager.getInstance().addSignOutEventListener(() -> {
+                        // Clean cached clusters
+                        instance.setCachedClusters(instance.additionalClusterDetails);
```
t-rufang: Are the linked Esp clusters still work if we sign out and sign in with another Azure account?

wezhang: The linked cluster are all in `additionalClusterDetails`

PR #3939, opened 2020-01-16
Enable AuthType auto selection when linking a HDInsight cluster

wezhang:


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/serverexplore/ui/AddNewClusterForm.java
```diff
@@ -313,6 +307,31 @@ protected void createUIComponents() {
...
+
+                if (StringUtils.isBlank(clusterNameOrUrl)) {
+                    return;
+                }
+
+                try {
+                    AuthType authType = SparkBatchSubmission.getInstance().probeAuthType(
+                            getClusterConnectionUrl(clusterNameOrUrl.trim()));
+
+                    authComboBox.setSelectedItem(authType);
```
t-rufang: Could you do some tests to determine whether `probeAuthType` is a time consuming method and cause UI stuck?

t-rufang: What if the authType is `AuthType.NotSupported`

wezhang: In my test, nothing will happen if the type isn't matched.

wezhang: About 500ms, you can feel that. 😄

PR #3944, opened 2020-01-17
Fix SparkBatchJobDebugExecutor equals() stack overflow

wezhang:
The SparkBatchJobDebugExecutor is registered into system components, whose
equal() will be called when getting components from ServiceManager.getService.
So, check the type at beginning to avoid the infinite loop by
DefaultDebugExecutor.getDebugExecutorInstance()


wezhang:
To address #3887

Issue #3921, opened 2020-01-13
[IntelliJ][ReportedByUser] Uncaught Exception  java.lang.Throwable: HIDE_DEFAULT_EDITOR is supported only for DumbAware providers; class com.microsoft.azure.hdinsight.jobs.framework.JobViewEditorProvider is not DumbAware.


jonmclean:
IntelliJ build version: 2019.3.1 IC-193.5662.53
OS: Linux
JDK: JetBrains s.r.o 11.0.5
Plugin version: 3.31.0-2019.3
Additional Info: Trying to open the Jobs view for an HDInsight cluster
Parent component: 
```
com.intellij.openapi.wm.impl.IdeRootPane[,0,25,2560x1375,invalid,layout=com.intellij.openapi.wm.impl.IdeRootPane$MyRootLayout,alignmentX=0.0,alignmentY=0.0,border=,flags=449,maximumSize=,minimumSize=,preferredSize=]
```
Error message: 
```
IdeaLoggingEvent[message=null, throwable=java.lang.Throwable: HIDE_DEFAULT_EDITOR is supported only for DumbAware providers; class com.microsoft.azure.hdinsight.jobs.framework.JobViewEditorProvider is not DumbAware.
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:145)
	at com.microsoft.azure.hdinsight.common.HDInsightHelperImpl$1.run(HDInsightHelperImpl.java:116)
```

jonmclean:
I received this error while trying to view HDInsight Spark jobs through the Azure plugin for IntelliJ Community Edition.

1. Open IntelliJ Community Edition
2. Open "Azure Explorer" plugin
3. Select Azure / HDInsight from the tree in the Azure Explorer
4. Select my HDInsight cluster from the tree
5. Select the "Jobs" folder from the tree.
6. Notice the Job View opens but is empty.
7. Notice the red flashing error symbol on the bottom right of the IDE. Click it.  A dialog box opens with the error displayed.
![AzureToolkitError](https://user-images.githubusercontent.com/4429525/72296042-9cb7fe00-360d-11ea-8cb9-9a3559f1ece2.png)

wezhang:
Thanks @jonmclean ! I'll check it.

jonmclean:
I verified that the issue also exists in the latest toolkit plugin version (3.31.0-2019.3).

jonmclean:
[...]

PR #3951, opened 2020-01-19
Fix HIDE_DEFAULT_EDITOR is only for DeumAware issue

Flanker32:
Add missing file in #3946 for HIDE_DEFAULT_EDITOR is only for DumbAware issue.


wezhang:
Addressed #3921

wezhang:
Addressed #3954

PR #3975, opened 2020-02-11
Add sub function node in explorer

Flanker32:
Add sub function node in explorer


Utils/azuretools-core/src/com/microsoft/azuretools/telemetry/TelemetryConstants.java
```diff
@@ -66,6 +67,9 @@
     public static final String RESTART_WEBAPP = "restart-webapp";
     public static final String WEBAPP_SHOWPROP = "showprop-webapp";
     public static final String WEBAPP_OPEN_INBROWSER = "open-inbrowser-webapp";
+    public static final String TRIGGER_FUNCTION = "function-trigger";
```
andxu: trigger-function

Flanker32: Fixed

PR #4006, opened 2020-02-21
replace depreacted log4j 1.2.17 with log4j-core

andxu:


andxu:
Close since log4j v1 and v2 are not compatible

PR #4064, opened 2020-03-01
1. add template for each trigger

andxu:
2. check mvn before creating function project.

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/resources/azurefunction/templates/ServiceBusTopicTrigger.template
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/resources/azurefunction/templates/ServiceBusTopicTrigger.template
@@ -1,0 +1,27 @@
+package $packageName$;
+
+import com.microsoft.azure.functions.annotation.*;
+import com.microsoft.azure.functions.*;
+
+/**
+ * Azure Functions with Service Topic Trigger.
+ */
+public class $className$ {
+    /**
+     * This function will be invoked when a new message is received at the Service Bus Topic.
+     */
+    @FunctionName("ServiceBusTopicTrigger-Java")
+    public void run(
+        @ServiceBusTopicTrigger(
+            name = "message",
+            topicName = "mysbtopic",
+            subscriptionName = "mysubscription",
+            connection = "MyStorageConnectionAppSetting"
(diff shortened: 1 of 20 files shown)


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/util/AzureFunctionsUtils.java
```diff
@@ -83,15 +98,26 @@ public static File createMavenProjectToTempFolder(final String groupId, final St
...
+                    .map(e -> e.getKey() + "=" + e.getValue())
+                    .collect(Collectors.joining(" "));
             CommandUtils.executeMultipleLineOutput(mvnCmd.getAbsolutePath() + " archetype:generate" + " -B " + args, folder);
 
             return folder;
         }
         return null;
     }
 
+    public static String generateFunctionClassByTrigger(String trigger, String packageName, String className) throws IOException {
```
Flanker32: Shall we keep a constant map including the default parameters instead of template files? Or we will have two copies of function template, in IntelliJ and in tools-common, besides, shall we move the default constants to `azure-tools-common` ;ater?

PluginsAndFeatures/azure-toolkit-for-intellij/resources/azurefunction/templates/ServiceBusTopicTrigger.template
```diff
@@ -0,0 +1,27 @@
...
+            subscriptionName = "mysubscription",
+            connection = "MyStorageConnectionAppSetting"
+        )
+        String message,
+        final ExecutionContext context
+    ) {
+        context.getLogger().info("Java Service Bus Topic trigger function executed.");
+        context.getLogger().info(message);
+    }
+}
```
Flanker32: New line at the end

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/util/AzureFunctionsUtils.java
andxu: we need to abandon mvn archetype since it is very slow on fresh machine

PR #4093, opened 2020-03-06
Trigger on select event after fill comboBox

Flanker32:
Trigger on select event after fill comboBox


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/forms/function/CreateFunctionForm.java
```diff
@@ -206,6 +194,33 @@ public void popupMenuWillBecomeInvisible(PopupMenuEvent e) {
         this.txtPackageName.setText(packageName);
     }
 
+    private void onSelectTriggerType() {
```
andxu: move private methods to the last part

PR #4103, opened 2020-03-10
Prepare for IntelliJ EAP 2020.1

t-rufang:
This is the MVP version for EAP 2020.1. With this PR we can install our plugin on IntelliJ EAP 2020.1 with no error. But there are still other efforts we need to put on EAP. For example, support dynamic loading, fix deprecate attribute in plugin.xml and fix deprecate API in codes.


BuildDevint
```diff
@@ -17,12 +17,12 @@ BUILD_ECLIPSE=true
 BUILD_INTELLIJ=true
 INJECT_INTELLIJ_VERSION=false
 
-IJ_VERSION_MAJOR_BUILD=193.5233
-IJ_VERSION_MAJOR=2019.3
+IJ_VERSION_MAJOR_BUILD=201.3803.71
+IJ_VERSION_MAJOR=LATEST-EAP-SNAPSHOT
```
wezhang: The `\gradle.propertie` also needs to change.

t-rufang: It's already changed in this PR. Check the source code at https://github.com/t-rufang/azure-tools-for-java/blob/eap201mvp/gradle.properties

PR #4126, opened 2020-03-16
Remove switch for Synapse feature in IntelliJ

t-rufang:


t-rufang:
> Thanks Rui! Could you help to remove `PluginsAndFeatures/azure-toolkit-for-intellij/resources/bundle/AadProvider.json`?

Fixed in new commit.

Issue #4107, opened 2020-03-11
[IntelliJ][ReportedByUser][2020.1 EAP] When create a new project, show this Uncaught Exception  java.lang.Throwable: Cannot add an action twice: Publish Group (Publish Group)


v-lanjli:
**IntelliJ build version:** 2020.1 EAP IC-201.6073.9
**OS:** Windows 10
**JDK:** JetBrains s.r.o 11.0.6
**Plugin version:** https://dev.azure.com/mseng/VSJava/_build/results?buildId=11679560&view=artifacts&type=publishedArtifacts
**Additional Info:** None

**Repor Step:**
Create a new project in an already opened project

**Parent component:** 
```
com.intellij.openapi.wm.impl.IdeRootPane[,8,8,1601x860,layout=com.intellij.openapi.wm.impl.IdeRootPane$MyRootLayout,alignmentX=0.0,alignmentY=0.0,border=,flags=449,maximumSize=,minimumSize=,preferredSize=]
```
**Error message:** 
```
IdeaLoggingEvent[message=null, throwable=java.lang.Throwable: Cannot add an action twice: Publish Group (Publish Group)
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:146)
	at com.microsoft.intellij.AzurePlugin.initializeAzureActionGroup(AzurePlugin.java:164)
	at com.microsoft.intellij.AzurePlugin.initComponent(AzurePlugin.java:148)
```

t-rufang:
Fixed in #4148

v-lanjli:
Verified and fixed in this build (https://dev.azure.com/mseng/VSJava/_build/results?buildId=11760285&view=artifacts&type=publishedArtifacts)
Closed this issue.

PR #4148, opened 2020-03-19
Fix add an action twice issue in EAP 2020.1

t-rufang:
- To address #4107

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/resources/META-INF/plugin.xml
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/resources/META-INF/plugin.xml
@@ -126,26 +126,14 @@
       implementationClass="com.microsoft.intellij.language.arm.parser.ARMParserDefinition"/>
     <completion.contributor language="JSON" order="first"
       implementationClass="com.microsoft.intellij.language.arm.codeinsight.ARMCompletionContributor"/>
+    <postStartupActivity implementation="com.microsoft.intellij.AzurePlugin"/>
   </extensions>
 
-  <application-components>
-    <component>
-      <implementation-class>com.microsoft.azure.hdinsight.jobs.JobViewApplicationComponent</implementation-class>
-    </component>
-    <component>
-      <implementation-class>com.microsoft.intellij.AzureActionsComponent</implementation-class>
-    </component>
-    <component>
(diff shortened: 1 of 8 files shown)


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/AzurePlugin.java
```diff
@@ -101,39 +98,24 @@
...
+    private AzureSettings azureSettings;
 
     private String installationID;
 
     private Boolean firstInstallationByVersion;
 
-    public AzurePlugin(Project project) {
-        super(project);
+    @Override
+    public void runActivity(@NotNull Project project) {
```
wezhang: As far as I know, in UT mode, the project might be `null`

t-rufang: This method is override from base class, please check the definition of the base class.
```java
public interface StartupActivity {
  ExtensionPointName<StartupActivity> POST_STARTUP_ACTIVITY = ExtensionPointName.create("com.intellij.postStartupActivity");

  /**
   * Please see https://github.com/JetBrains/intellij-community/blob/master/platform/service-container/overview.md#startup-activity
   */
  ExtensionPointName<StartupActivity.Background> BACKGROUND_POST_STARTUP_ACTIVITY = ExtensionPointName.create("com.intellij.backgroundPostStartupActivity");

  void runActivity(@NotNull Project project);

  interface DumbAware extends StartupActivity, com.intellij.openapi.project.DumbAware {
  }

  interface Background extends StartupActivity, com.intellij.openapi.project.DumbAware {}
}
```

PluginsAndFeatures/azure-toolkit-for-intellij/resources/META-INF/plugin.xml
```diff
@@ -126,26 +126,14 @@
       implementationClass="com.microsoft.intellij.language.arm.parser.ARMParserDefinition"/>
     <completion.contributor language="JSON" order="first"
       implementationClass="com.microsoft.intellij.language.arm.codeinsight.ARMCompletionContributor"/>
+    <postStartupActivity implementation="com.microsoft.intellij.AzurePlugin"/>
```
wangmingliang-ms, line 129: why use `postStartupActivity` but not `startupActivity`?

PR #4168, opened 2020-03-22
fix line end and java header

andxu:


PluginsAndFeatures/AddLibrary/AppInsights/com.microsoft.applicationinsights.ui/src/main/java/com/microsoft/applicationinsights/preference/ApplicationInsightsAddDialog.java
```diff
@@ -19,7 +19,6 @@
  * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-
```
wezhang: Looks like we can keep this line break

andxu:
replaced with https://github.com/microsoft/azure-tools-for-java/pull/4169

Issue #4213, opened 2020-03-29
[Azure Functions] unhandled error when creating function app


yucwan:
If you click ok without app service plan selected:
![image](https://user-images.githubusercontent.com/49381248/77852859-73257080-7213-11ea-8401-948057895238.png)

it will throw an unhandled error:
![image](https://user-images.githubusercontent.com/49381248/77852850-69037200-7213-11ea-992e-fad4ba64197a.png)

Flanker32:
Fixed in #4239

PR #4239, opened 2020-04-07
Update validation in function creation dialog

Flanker32:
Use AzureDialogWrapper to fix function creation dialog validation issues, for #4213


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/functions/deploy/ui/creation/FunctionCreationDialog.java
```diff
@@ -249,39 +179,39 @@ private void updateConfiguration() {
         return appSettings;
     }
 
-    private boolean validateConfiguration() {
+    @Override
+    protected List<ValidationInfo> doValidateAll() {
         updateConfiguration();
```
andxu, line 184: no need to updateConfiguration

PR #4247, opened 2020-04-08
Fix the issue: missing <Empty> options after switching clusters.

andxu:


Flanker32:
Duplicated with #4248

PR #4248, opened 2020-04-08
Mark border when the value is dirty

andxu:


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/helpers/springcloud/SpringCloudAppPropertyView.java
```diff
@@ -343,12 +345,38 @@ private void syncSaveStatus() {
...
+
+    private void removeHighLight(final JComponent component) {
+        Border border = borderMap.remove(component);
+        if (border != null) {
+            component.setBorder(border);
+        }
+    }
+
+    private void addHighLight(final JComponent component) {
+        borderMap.putIfAbsent(component, component.getBorder());
```
Flanker32, line 376: Why we use a map here. seems a List is enough.
Besides, ConcurrentHashMap will throw exception when null values, this may occurs for Jcomponent

andxu: Fixed

andxu:
Replaced with PR https://github.com/microsoft/azure-tools-for-java/pull/4250

PR #4274, opened 2020-04-10
Support trigger function event hub trigger

Flanker32:
Support trigger function event hub trigger


Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/function/SubFunctionNode.java
```diff
@@ -124,23 +123,48 @@ private void triggerTimerTrigger() {
...
-        JsonObject jsonObject = new Gson().fromJson(new InputStreamReader(response.getEntity().getContent()),
-                JsonObject.class);
-        return jsonObject.get("masterKey").getAsString();
+        request.setHeader("x-functions-key", masterKey);
+        request.setHeader("Content-Type", "application/json");
+        return request;
+    }
+
+    // work around for API getMasterKey failed
+    private synchronized String getFunctionMasterKey() throws IOException {
```
andxu, line 149: move this method to parent node

Flanker32: Fixed

PR #4329, opened 2020-04-23
Fix typo

pablomh:
occured -> occurred.

Found while testing via Terraform :)


andxu:
Thank @pablomh, the typo fix is approved.

Issue #4016, opened 2020-02-24
Weak Cryptographic Hash


QiAnXinCodeSafe:
https://github.com/microsoft/azure-tools-for-java/blob/d599b052bb809fdf331ed1e3975d3c5511ed7367/Utils/azuretools-core/src/com/microsoft/azuretools/azurecommons/util/CerPfxUtil.java#L126

Weak cryptographic hashes cannot guarantee data integrity and should not be used in security-critical contexts.

PR #4462, opened 2020-06-18
Remove useless code to fix issue #4016: Weak Cryptographic Hash

andxu:


andxu:
Fix #4016

andxu:
Cannot remove this class since there is an action in eclipse

andxu:
Replace this PR with https://github.com/microsoft/azure-tools-for-java/pull/4464

PR #4572, opened 2020-08-25
[github #1125][devops #1764066] Update apiVersion in Intellij Plugin for Azure Spring Cloud

wangmingliang-ms:
do same thing as https://github.com/microsoft/azure-maven-plugins/pull/1126


wangmingliang-ms:
result of `Enable`/`Disable` [Public Endpoints]/[Persistent Storage] can not be reflected immediately after configuration is saved(user have to wait some seconds and then refresh to check the reflected update)

andxu (dismissed):

PR #4588, opened 2020-09-09
Remove docker host codes and outdated key vault library

Flanker32:
Remove docker host codes and outdated key vault library, fixes https://msazure.visualstudio.com/Unified%20Platform%20KPIs/_workitems/edit/6478137


wezhang (approved):
LGTM! Shipit!

andxu (approved):

PR #4595, opened 2020-09-17
Task 1772676: Azure Toolkit for IntelliJ/Eclipse - Docker Run/Run on Web App for Containers ... Throws javax.ws.rs.ProcessingException

wangmingliang-ms:


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/container/utils/DockerUtil.java
```diff
@@ -50,6 +52,8 @@
 
 
 public class DockerUtil {
```
andxu, line 54: why there are two same DockerUtil?

wangmingliang-ms: they are for eclipse/intellij separately.
they should be merged and shared. there are many other such kind of duplicate classes in eclipse/intellij.
we can refactor in the future.

andxu: ok

andxu (approved):

PR #4596, opened 2020-09-17
add ssh into web app function.

shenqianjin:


Utils/azuretools-core/src/com/microsoft/azuretools/utils/CommandUtils.java
```diff
@@ -149,4 +178,47 @@ private static String getSafeWorkingDirectory() {
...
+        public String getOutputMessage() {
+            return outputMessage;
+        }
+
+        public void setOutputMessage(String outputMessage) {
+            this.outputMessage = outputMessage;
+        }
+    }
+
+    private static class CommendExecution {
```
wangmingliang-ms, line 203: is it possible not to create new concepts/classes

PR #4602, opened 2020-09-21
fixed the issue left click doesn't work for tree node.

shenqianjin:


Flanker32 (approved):

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/components/ServerExplorerToolWindowFactory.java
```diff
@@ -205,6 +204,15 @@ private void treeMousePressed(MouseEvent e, JTree tree) {
         }
     }
 
+    private Node getTreeNodeOnMouseClick(JTree tree, TreePath treePath) {
```
andxu, line 207: static?

shenqianjin: the class is used with its instance, no need to make it as static.

Issue #4601, opened 2020-09-21
[IntelliJ][Spark] Spark job configuration is overridden by blob uploading seeting


wezhang:
Plugin version: 3.42

When selects Blob uploading and sets Job Configuration as below:
![image](https://user-images.githubusercontent.com/7385667/93728810-45364d00-fbf4-11ea-852f-d1cb1cd7fc35.png)

But only has the following conf in submission:

    "conf": {
        "spark.hadoop.fs.azure.account.key.xxxx.blob.core.windows.net": "..."
    },

Other spark.yarn... configurations are missing

wezhang:
@v-lanjli @v-xiaocw

t-rufang:
@v-lanjli This bug is fixed in #4603. Please help verify the fix.

v-lanjli:
Verified and fixed. Closed this issue.

PR #4603, opened 2020-09-21
Fix config is  wrongly cleared after blob storage is slected as Synapse batch job issue

t-rufang:
- To address #4601


wezhang (approved):
LGTM! Thanks Rui! Shipit!

PR #4623, opened 2020-09-27
Fix null tenant id issues

Flanker32:
- Initialize subscription manager while restore intellij credential
- Cache all subscription sid -> tid relation ship instead of the active subscriptions in subscription manager
- Clear AzureCliAzureManager token cache when sign out

Fixes https://dev.azure.com/mseng/VSJava/_workitems/edit/1775839


andxu:
AB#1775839

andxu (approved):

wangmingliang-ms (approved):

PR #4685, opened 2020-10-27
Add function creation dialog and related components

Flanker32:
Add function creation dialog and related components

**Basic Mode(Same as web app)**
image

**Advance Mode**
image

image

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/appservice/AppServiceConfigFormPanelAdvanced.java
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/appservice/AppServiceConfigFormPanelAdvanced.java
@@ -36,9 +36,9 @@ import com.microsoft.azure.toolkit.intellij.appservice.serviceplan.ServicePlanCo
 import com.microsoft.azure.toolkit.intellij.appservice.subscription.SubscriptionComboBox;
 import com.microsoft.azure.toolkit.intellij.common.AzureArtifactComboBox;
 import com.microsoft.azure.toolkit.intellij.common.AzureFormPanel;
+import com.microsoft.azure.toolkit.lib.appservice.AppServiceConfig;
 import com.microsoft.azure.toolkit.lib.appservice.Platform;
 import com.microsoft.azure.toolkit.lib.common.form.AzureFormInput;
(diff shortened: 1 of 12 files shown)


shenqianjin (commented):
looks good!

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/appservice/AppServiceConfigFormPanelAdvanced.java
```diff
@@ -74,15 +76,16 @@
     private TitledSeparator deploymentTitle;
     private JLabel deploymentLabel;
 
-    public AppServiceConfigFormPanelAdvanced(final Project project) {
+    public AppServiceConfigFormPanelAdvanced(final Project project, final Supplier<T> supplier) {
```
wangmingliang-ms, line 79: is supplier a required part of `AppServiceConfigFormPanelAdvanced`? otherwise IMO, it should not be a constructor parameter

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/appservice/AppServiceConfigFormPanelAdvanced.java
```diff
@@ -142,6 +145,10 @@ public void setDeploymentVisible(boolean visible) {
         this.selectorApplication.setVisible(visible);
     }
 
+    public SubscriptionComboBox getSelectorSubscription() {
+        return selectorSubscription;
+    }
```
wangmingliang-ms, line 150: it's NG to make it public.
I would prefer making `selectorSubscription` `protected` and then override the `init` method to add another listener.

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/appservice/AppServiceConfigFormPanelAdvanced.java
Flanker32: @wangmingliang-ms Yes, we need to use the `supplier` to build the Config object

wangmingliang-ms (approved):

PR #4694, opened 2020-10-30
process special charater on openning web app in brower after deploying successfully.

shenqianjin:


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/webapp/webappconfig/WebAppRunState.java
```diff
@@ -212,7 +212,7 @@ private String getUrl(@NotNull WebAppBase webApp, @NotNull String fileName, @Not
         String url = "https://" + webApp.defaultHostName();
         if (Comparing.equal(fileType, MavenConstants.TYPE_WAR)
             && !webAppSettingModel.isDeployToRoot()) {
-            url += "/" + fileName;
+            url += "/" + WebAppUtils.encodeURL(fileName).replaceAll("\\+", "%20");
```
andxu, line 215: why do we need to replace '+' to %20?

shenqianjin: These rules are quite almost the same, except for the treatment of the space character. Per HTML 4, space should be encoded into + where as RFC-3986 requires it to be encoded into %20.

andxu (approved):

PR #4707, opened 2020-11-03
Bug 1774224: User should be prompt with the right message when tokens in azure cli expires.

wangmingliang-ms:
… in azure cli expires.


Utils/azuretools-core/src/com/microsoft/azuretools/enums/ErrorEnum.java
```diff
@@ -28,6 +28,8 @@
 public enum ErrorEnum {
     UNKNOWN_HOST_EXCEPTION(100000, "Encountered an unknown host exception.",
             "It seems you have an unstable network at the moment, please try again when network is available."),
+    INVALID_AUTHENTICATION(100401, "Invalid authentication",
+            "It seems you are not signed in properly, try signing out and then signing in again later"),
```
andxu, line 32: the message is confusing when using azure cli, user need to sign-out at az, otherwise signing out and then signing in again will not solve this problem.

wangmingliang-ms: let's ask PM for advice.

andxu (approved):

PR #4711, opened 2020-11-04
Fix function deployment id storage issue

Flanker32:
- Save created function app id to function deployment configuration
- Do not save empty `AZURE_WEB_JOB_STORAGE_KEY` to deployment configuration as they will be created by SDK automatically


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/runner/functions/component/table/AppSettingsTable.java
```diff
@@ -70,14 +69,18 @@ public void removeAppSettings(int row) {
         this.refresh();
     }
```
andxu, line 71: what's the meaning of clearRequiredAttributes

andxu (approved):

PR #4738, opened 2020-11-10
Record all unhandled exceptions and restful API exceptions 

Flanker32:
- Send telemetry for unhandled exception
*- Track rest exception in SDK call(Removed)*


shenqianjin (commented):
Can we confirm that every exception relate to rest such as TimeoutExcetion/UnknownHostException has already been correctly converted to com.microsoft.rest.RestException ?
If we can, it should be ok. Or, maybe we should include some other exceptions of rest invoking into the try-catch block.

Flanker32:
> Can we confirm that every exception relate to rest such as TimeoutExcetion/UnknownHostException has already been correctly converted to com.microsoft.rest.RestException ?
> If we can, it should be ok. Or, maybe we should include some other exceptions of rest invoking into the try-catch block.

Checked the SDK behavior, RestException by rest call result will not be catched in `TelemetryInterceptor`, so revert the commit

wangmingliang-ms (approved):

PR #4742, opened 2020-11-12
Task 1792621: [refactor] IOException is not used properly and severely polluted our code

wangmingliang-ms:
Almost all the fixed IOException are thrown from auth related code. 
User must signin first before they can use any feature of the plugin, so we can just throw a runtime exception to terminate user action if auth related problem found.

these are the original methods throwing IOException before fix (which finally pollutes >50 classes and hundreds of methods) 
image

### DIFF
--- a/Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/Node.java
+++ b/Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/Node.java
@@ -22,11 +22,9 @@
 
 package com.microsoft.tooling.msservices.serviceexplorer;
 
-import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterators;
 import com.microsoft.azuretools.authmanage.AuthMethodManager;
-import com.microsoft.azuretools.azurecommons.helpers.AzureCmdException;
 import com.microsoft.azuretools.azurecommons.helpers.NotNull;
 import com.microsoft.azuretools.azurecommons.helpers.Nullable;
 import com.microsoft.azuretools.core.mvp.ui.base.MvpView;
@@ -42,7 +40,6 @@ import javax.swing.*;
 import javax.swing.tree.TreePath;
 import java.beans.PropertyChangeListener;
 import java.beans.PropertyChangeSupport;
-import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
@@ -138,11 +135,13 @@ public class Node implements MvpView, BasicTelemetryProperty {
     }
 
     public boolean isDescendant(Node node) {
-        if (isDirectChild(node))
+        if (isDirectChild(node)) {
             return true;
-        for (Node child : childNodes) {
-            if (child.isDescendant(node))
+        }
+        for (final Node child : childNodes) {
+            if (child.isDescendant(node)) {
                 return true;
+            }
         }
 
         return false;
@@ -150,11 +149,13 @@ public class Node implements MvpView, BasicTelemetryProperty {
 
     // Walk up the tree till we find a parent node who's type
     // is equal to "clazz".
-    public <T extends Node> T findParentByType(Class<T> clazz) {
-        if (parent == null)
+    public <T extends Node> @Nullable T findParentByType(Class<T> clazz) {
+        if (parent == null) {
             return null;
-        if (parent.getClass().equals(clazz))
+        }
+        if (parent.getClass().equals(clazz)) {
             return (T) parent;
+        }
         return parent.findParentByType(clazz);
     }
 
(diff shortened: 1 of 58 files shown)


Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/Node.java
```diff
@@ -328,12 +309,7 @@ public void onNodeDblClicked(Object context) {
     }
 
     public NodeAction getNodeActionByName(final String name) {
-        return Iterators.tryFind(nodeActions.iterator(), new Predicate<NodeAction>() {
-            @Override
-            public boolean apply(NodeAction nodeAction) {
-                return name.compareTo(nodeAction.getName()) == 0;
-            }
-        }).orNull();
+        return Iterators.tryFind(nodeActions.iterator(), nodeAction -> name.compareTo(nodeAction.getName()) == 0).orNull();
```
Flanker32, line 312: use `StringUtils.equals` ? As `compareTo` may more likely a sort method

Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/exception/AzureToolkitRuntimeException.java
```diff
@@ -0,0 +1,47 @@
...
+
+package com.microsoft.azure.toolkit.lib.common.exception;
+
+import lombok.Getter;
+
+@Getter
+public class AzureToolkitRuntimeException extends RuntimeException {
+    private final String action;
+
+    public AzureToolkitRuntimeException(String error) {
```
Flanker32, line 31: error->message?

Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/AuthMethodManager.java
```diff
@@ -68,34 +64,36 @@ private AuthMethodManager(AuthMethodDetails authMethodDetails) {
         this.authMethodDetails = authMethodDetails;
         // initialize subscription manager when restore authentication
         if (this.authMethodDetails.getAuthMethod() != null) {
-            try {
-                getAzureManager().getSubscriptionManager().updateSubscriptionDetailsIfNull();
-            } catch (IOException e) {
-                // swallow exception
-            }
+            Objects.requireNonNull(getAzureManager()).getSubscriptionManager().updateSubscriptionDetailsIfNull();
```
Flanker32, line 67: I think we just need to check `getAzureManager() == null` and return, the constructor will be called in `LazyLoader` during plugin initialization, throw NPE here may failed the process.

Flanker32 (changes requested):

Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/AuthMethodManager.java
```diff
@@ -68,34 +64,36 @@ private AuthMethodManager(AuthMethodDetails authMethodDetails) {
...
-    public AppPlatformManager getAzureSpringCloudClient(String sid) throws IOException {
-        if (getAzureManager() == null) {
-            throw new IOException(CANNOT_GET_AZURE_MANAGER);
+    public AppPlatformManager getAzureSpringCloudClient(String sid) {
+        final AzureManager manager = getAzureManager();
+        if (manager != null) {
+            return getAzureManager().getAzureSpringCloudClient(sid);
         }
-        return getAzureManager().getAzureSpringCloudClient(sid);
+        final String error = "Failed to initialize request";
```
Flanker32, line 94: duplicate with previous, create an exception `AzureToolkitNotSignedInException`?

Utils/azuretools-core/src/com/microsoft/azuretools/core/mvp/model/function/AzureFunctionMvpModel.java
```diff
@@ -66,40 +53,42 @@ public static AzureFunctionMvpModel getInstance() {
...
 
-    public FunctionApp getFunctionById(String sid, String id) throws IOException {
-        FunctionApp app = getFunctionAppsClient(sid).getById(id);
-        if (app == null) {
-            throw new IOException(CANNOT_GET_FUNCTION_APP_WITH_ID + id); // TODO: specify the type of exception.
+    @NotNull
+    public FunctionApp getFunctionById(String sid, String id) throws AzureToolkitRuntimeException {
+        final FunctionApp app = getFunctionAppsClient(sid).getById(id);
+        if (Objects.isNull(app)) {
+            final String error = String.format("cannot find FunctionApp[%s] in subscription[%s]", id, sid);
```
andxu, line 60: Cannot
Confirm

andxu (dismissed):

Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/AuthMethodManager.java
wangmingliang-ms: let's summarize such kind of common exceptions in the future.

PR #4744, opened 2020-11-13
Task 1793111: menu is not shown immediately(blocked) on right click on web app node.

wangmingliang-ms:


andxu (approved):

Flanker32 (approved):

Flanker32 (approved):
We may need to move the enable statement to correspond action class as flight recorder action was not added in WebAppNode

PR #4770, opened 2020-11-25
Task 1796326: id is too long to show on error dialog

wangmingliang-ms:


Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/operation/AzureOperationUtils.java
```diff
@@ -46,7 +47,15 @@ public static String getOperationTitle(@NotNull AzureOperationRef ref) {
         final AzureOperation annotation = AzureOperationUtils.getAnnotation(ref);
         final String messageTemplate = annotation.value();
         final String[] parameters = annotation.params();
-        final String[] params = Arrays.stream(parameters).map(expression -> interpretExpression(expression, ref)).toArray(String[]::new);
+        final String[] params = Arrays.stream(parameters).map(expression -> {
+            final String str = interpretExpression(expression, ref);
+            //TODO: @wangmi improve this part to get name from uri
+            if (Objects.nonNull(str) && str.contains("/")) {
+                final String[] parts = str.split("/");
+                return str.endsWith("/") ? parts[parts.length - 2] : parts[parts.length - 1];
```
andxu, line 55: test "a/"

wangmingliang-ms: would return `a`

PR #4796, opened 2020-12-04
Improve "Trigger Function" logic for httpTrigger functions

maartenba:
This PR fixes two issues when triggering HTTP functions from Azure Explorer:

* Correctly detect HTTP trigger functions, as they do not have `"direction": "in"` in the function envelope
* Fix parsing enum value for HTTP trigger authorization (it may be lowercase in the envelope)


andxu:
LGTM, add @Flanker32 to confirm

Flanker32:
@maartenba Thanks a lot for your efforts, though http trigger with no direction could works, but from the function trigger [document](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-trigger?tabs=java#configuration), direction is required and must be `in`. 
By the way, it seems that we could skip the direction test as there should be only one trigger in each function and direction will always be `in`, for details, please refer document [here](https://docs.microsoft.com/en-us/azure/azure-functions/functions-triggers-bindings?tabs=csharp#binding-direction)

maartenba:
@Flanker32 I ran into this when using the Azure Toolkit on a .NET deployed function. It seems that SDK omits "direction in", which renders the trigger action useless for those. Not sure about other runtimes though.

Want me to update and remove the direction check?

Flanker32:
> @Flanker32 I ran into this when using the Azure Toolkit on a .NET deployed function. It seems that SDK omits "direction in", which renders the trigger action useless for those. Not sure about other runtimes though.
> 
> Want me to update and remove the direction check?

Yes, I think we could just remove the direction check, just ensure the binding type is trigger

Flanker32 (approved):
LGTM, Thanks!

PR #4813, opened 2020-12-15
Upgrade azure blob batch sdk to 12.7.0 with excluded in spark local r…

wezhang:
…un mock

The upgraded netty-buffer 4.1.52 removed symbols depends by Spark


t-rufang (approved):
Thanks Wei! LGTM!

PR #4828, opened 2020-12-17
Revert "Create runIdeU task for IDEA Ultimate development"

wezhang:
This reverts commit 1d18cfcd, which has side impacts to `runIde`


t-rufang (approved):
Tks Wei!

PR #4829, opened 2020-12-17
Fix UI Blocking when restoring Gen2 storage key

wezhang:


t-rufang (approved):
Thanks Wei!

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/hdinsight/spark/ui/SparkSubmissionJobUploadStorageGen2Card.kt
```diff
@@ -106,7 +108,10 @@ class SparkSubmissionJobUploadStorageGen2Card : SparkSubmissionJobUploadStorageB
                 if (!storageKeyField.text.isNullOrBlank()) {
                     val viewModel = viewModel as? ViewModel ?: return
                     val credentialAccount = ADLS_GEN2.getSecureStoreServiceOf(viewModel.rootUri?.accountName) ?: return
-                    storageKeyField.text = secureStore?.loadPassword(credentialAccount, gen2Account) ?: ""
+
+                    Observable.just(secureStore?.loadPassword(credentialAccount, gen2Account) ?: "")
+                            .observeOn(IdeaSchedulers().dispatchUIThread())
+                            .subscribe { storageKeyField.text = it }
```
t-rufang, line 114: Maybe we need to handle errors in this observable?

wezhang: It's supposed not throwing any exceptions here. For the runtime throwable, should be a bug to report to user.

PR #4838, opened 2020-12-22
[Azure MySQL] switch MySQL SDK to release version

shenqianjin:


Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/mysql/MySQLNode.java
```diff
@@ -106,6 +107,21 @@ public DeleteAzureMySQLAction() {
                     String.format(ACTION_DOING + StringUtils.SPACE + MySQLModule.ACTION_PATTERN_SUFFIX, MySQLNode.this.name));
         }
 
+        @Override
+        public int getGroup() {
+            return super.getGroup() + AzureActionEnum.DELETE.getGroup();
```
Flanker32, line 112: why add `super.getGroup()`, instead of using DELETE group directly

shenqianjin: the group/priority in AzureActionEnum is a relative value.

shenqianjin: will refactor to enhance configuration about priority and group in the next sprint.

Flanker32 (approved):

PR #4847, opened 2020-12-24
Bug 1803723: [Bug][AzureToolkit][Test] Error box begin with failed to load/refresh items of combo box

wangmingliang-ms:
… load/refresh items of combo box

### DIFF
--- a/Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/handler/AzureExceptionHandler.java
+++ b/Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/handler/AzureExceptionHandler.java
@@ -22,10 +22,13 @@
 package com.microsoft.azure.toolkit.lib.common.handler;
 
 import com.microsoft.azuretools.azurecommons.helpers.Nullable;
+import lombok.extern.java.Log;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 
 import java.io.InterruptedIOException;
+import java.util.logging.Level;
 
+@Log
 public abstract class AzureExceptionHandler {
     private static AzureExceptionHandler handler;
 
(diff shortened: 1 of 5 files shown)


Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/operation/AzureOperationRef.java
```diff
@@ -23,15 +23,26 @@
...
     private Object instance;
+
+    @Override
+    public boolean equals(final Object obj) {
+        if (!(obj instanceof AzureOperationRef)) {
+            return false;
+        }
+        final AzureOperationRef operation = (AzureOperationRef) obj;
+        return operation.getMethod() == this.method || operation.getMethod().equals(this.method);
+    }
```
shenqianjin, line 47: only "operation.getMethod().equals(this.method)" should be enough. right ?

Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/handler/AzureExceptionHandler.java
```diff
@@ -53,11 +56,13 @@ public static void onRxException(final Throwable e) {
     }
 
     public void handleException(Throwable throwable, @Nullable AzureExceptionAction... action) {
+        log.log(Level.WARNING, "caught an error in AzureExceptionHandler", throwable);
```
shenqianjin, line 59: I prefer to log.log(Level.WARNING, throwable);

Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/operation/AzureOperationRef.java
wangmingliang-ms: yes

Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/handler/AzureExceptionHandler.java
wangmingliang-ms: so do I, but there is no such method.

PR #4853, opened 2020-12-28
User Story 1803685: refactor blocking progress window to non-blocking

wangmingliang-ms:


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/common/task/IntellijAzureTaskManager.java
```diff
@@ -22,63 +22,99 @@
...
     }
 
     @Override
-    protected void doRunInModal(final AzureTask task) {
+    protected void doRunInModal(final Runnable runnable, final AzureTask<?> task) {
+        if (task.isBackgroundable()) {
+            this.doRunInBackgroundableModal(runnable, task);
+        } else {
+            this.doRunInUnBackgroundableModal(runnable, task);
+        }
```
shenqianjin, line 81: Unbackgroundable ?
btw, I prefer to name it as "xxxConditionalModel", just like Intellij ConditionnalModel.class.

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/common/handler/IntelliJAzureExceptionHandler.java
```diff
@@ -211,6 +213,18 @@ private String getActionText(final List<Throwable> throwableList) {
         return registerAction == null ? actions : ArrayUtils.addAll(actions, registerAction);
     }
 
+    public static List<AzureOperationRef> revise(Deque<? extends AzureOperationRef> operations) {
+        final LinkedList<AzureOperationRef> result = new LinkedList<>();
+        for (final AzureOperationRef op : operations) {
```
Flanker32, line 218: So we only track operation stack to the latest action? What if user action call another action, like create web app will call refresh explorer, which may hide the origin user action.

Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/task/AzureOperationEnhancer.java
```diff
@@ -20,45 +20,38 @@
...
         final AzureOperationRef operation = toOperationRef(point);
-        if (CollectionUtils.isNotEmpty(AzureOperationsContext.getOperations())) {
-            final AzureOperationRef popped = AzureOperationsContext.pop();
-            if (!Objects.equals(popped, operation)) {
-                log.log(Level.SEVERE, "!!!operation stack is wrongly managed!!!");
-            }
-        }
+        log.info(String.format("exit operation[%s] in context[%s]", operation, AzureTaskContext.current()));
+        final AzureOperationRef popped = AzureTaskContext.current().popOperation();
+        assert Objects.equals(popped, operation) : String.format("popped operation[%s] is not the exiting operation[%s]", popped, operation);
```
Flanker32, line 54: should we break users operation when meet issues in operation tracking? I think `log.error` should be enough

wangmingliang-ms: it's is a fatal error if this assertion fails. going forward will just cause unexpected errors.

Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/task/AzureTask.java
```diff
@@ -22,39 +22,78 @@
...
+    private final boolean backgroundable = true;
+    private final String title;
+
+    @Setter(AccessLevel.PACKAGE)
+    private AzureTaskContext.Node context;
 
     public AzureTask(Runnable runnable) {
-        this.runnable = runnable;
-        this.modality = Modality.DEFAULT;
+        this(runnable, Modality.DEFAULT);
```
Flanker32, line 46: create a builder for `AzureTask`, as there are too many overloader for constructor which may be a bit confusing

Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/task/AzureOperationEnhancer.java
Flanker32: Operation annotation should not affect the logic, why will be fatal error here? Shall we catch exception here and log error?

PluginsAndFeatures/azure-toolkit-for-intellij/resources/com/microsoft/intellij/ui/messages/messages.properties
```diff
@@ -1488,7 +1488,9 @@ function.cli.error.notFound=Azure Functions Core Tools not found. Please go to h
 function.binding.error.parseFailed=Cannot parse annotation information, expect type: %s, actual type: %s
 function.binding.storage.found=StorageAccount annotation found.
 function.binding.storage.notFound=No StorageAccount annotation found.
-webapp.create.task.title=Creating New Web App...
+webapp.create.task.title=Creating new web app...
+webapp.create.task.success.notification.title=Create web app
+webapp.create.task.success.notification.message=Web App[%s] is successfully created.
```
shenqianjin, line 1493: use "()" rather than "[]" ?

Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/operation/AzureOperationRef.java
```diff
@@ -24,25 +24,30 @@
...
         final AzureOperationRef operation = (AzureOperationRef) obj;
-        return operation.getMethod() == this.method || operation.getMethod().equals(this.method);
+        return operation.getMethod() == this.method || Objects.equals(operation.getMethod(), this.getMethod());
+    }
+
+    @Override
+    public String toString() {
+        final AzureOperation annotation = AzureOperationUtils.getAnnotation(this);
+        return String.format("{title:'%s', method:%s}", annotation.value(), method.getName());
     }
```
shenqianjin, line 52: do we need to check whether annotation == null here?

Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/task/AzureOperationEnhancer.java
wangmingliang-ms: > Operation annotation should not affect the logic, why will be fatal error here?
IMO. the code should never come here, this assertion failure means our error handling framework has a error/bug, but not exception. and going forward is meaningless.
it's just similar to a framework's inner error or system error(e.g. `OutOfMemoryError`). which can not recover and the user can do nothing.

And we would expect the user report it to developer. 
As we have discussed, 
for `AzureToolkitException`/`AzureToolkitRuntimeException`, we will not show `report button`.
while for unknow exceptions(i.e. **non**-`AzureToolkitException`/`AzureToolkitRuntimeException`, including `AssertionException`/`NullPointerException`...), our root error handler will show a `report` button on the dialog/notification.

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/common/handler/IntelliJAzureExceptionHandler.java
wangmingliang-ms: This is a compromise solution to our current implementation, and I am actually considering if we should add another type of operation.

Think about the case: refreshing explorer fails after successfully create a web app. 
what should the message looks like? 
`failed to create web app, as refreshing explorer failed ?`
No, it's not the fact since webapp is actually successfully created.

the problem is that. some operations are required part of the parent operation, and the others are not(optional or just side effects) e.g. 
in our case/implementation of web app creation. 
```
USER ACTION(create web app) {
    1. SERVICE(check login status/login)
    2. "SERVICE(open dialog)"
      3. SERVICE(create web app)
        4. ACTION(deploy artifact)
        5. ACTION(refresh explorer)
}
```
1 & 2 & 3 are required parts of `create web app`, the failure of these operations will cause failure of `create webapp`.
but 4 & 5 are not the require parts, the failure of these operations will **NOT** cause failure of `create webapp`.

as a compromise to our current design/implementation, I make these optional/side effect operations `ACTION`.
I have come up with a very raw idea to avoid this kind of compromise. let's discuss it when you come back.

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/common/task/IntellijAzureTaskManager.java
wangmingliang-ms: this method is actually implemented based on Intellij's `Task.Modal`.
with `Unbackgroudable`, I just want to emphasize that it can not be sent to background.

I actually don't understand what `ConditionalModal` is used for. and we don't use it for now.

Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/operation/AzureOperationRef.java
wangmingliang-ms: no, the annotation will never be null, otherwise this `AzureOperationRef` instance will not be created.

Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/task/AzureTask.java
wangmingliang-ms: builder is valid but not sure if it's good here. 
since I want to emphasize that some parameters are required(e.g. the runnable/supplier).

andxu:
This PR is too large, please split it into small PRs.

PR #4856, opened 2020-12-30
[Icon Improvement] (No.4) Add generic actions icons on Function App - refactor telemetry constants

shenqianjin:


Utils/azuretools-core/src/com/microsoft/azuretools/telemetry/TelemetryParameter.java
```diff
@@ -0,0 +1,145 @@
...
+    @Override
+    public int hashCode() {
+        return this.operationName.hashCode();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        } else if (!(obj instanceof TelemetryParameter)) {
```
andxu, line 135: remove all else's

Utils/azuretools-core/src/com/microsoft/azuretools/telemetry/TelemetryParameter.java
```diff
@@ -0,0 +1,145 @@
...
+            return false;
+        } else if (obj == this) {
+            return true;
+        } else {
+            TelemetryParameter rhs = (TelemetryParameter) obj;
+            return this.operationName.equalsIgnoreCase(rhs.operationName);
```
andxu, line 141: use EqualsBuilder?

shenqianjin: equals is enough here. just need to compare String and the constructor is private for this class.

andxu (approved):

PR #4871, opened 2020-12-30
Non blocking/1 create webapp

wangmingliang-ms:


andxu (approved):

Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/task/AzureTask.java
```diff
@@ -22,15 +22,24 @@
...
     private Modality modality;
     private Runnable runnable;
     private Object project;
     private boolean cancellable;
+    @Builder.Default
+    private boolean backgroundable = true;
+    private boolean runningBackground = true;
     private String title;
+    private Runnable successListener;
+    private Runnable finishedListener;
```
andxu, line 41: finishedListener is not proper

PR #4894, opened 2021-01-15
[Action Refactor] (No.3) Remove dependency on Backgroundable/Promptable/Telemetrable on Function App

shenqianjin:
…le/Telemetrable on Function App

### DIFF
--- a/Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/BasicActionBuilder.java
+++ b/Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/BasicActionBuilder.java
@@ -26,6 +26,8 @@ import com.google.common.base.Preconditions;
 import com.microsoft.azuretools.telemetry.TelemetryParameter;
 import org.apache.commons.lang3.StringUtils;
 
+import java.util.Objects;
+
 public class BasicActionBuilder {
 
     private static final String FUZZY_PROGRESS_MESSAGE_PATTERN = "%s...";
@@ -36,6 +38,7 @@ public class BasicActionBuilder {
 
     private Runnable runnable;
     private AzureActionEnum action;
+    private String doingName;
 
     private String moduleName;
     private String instanceName;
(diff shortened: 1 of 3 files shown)


Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/BasicActionBuilder.java
```diff
@@ -36,6 +38,7 @@
 
     private Runnable runnable;
     private AzureActionEnum action;
+    private String doingName;
```
Flanker32, line 41: Rename this parameter? like `actionName` ?

shenqianjin: It is not an Action Name, it is a action DOING name such as starting/stopping/etc.

Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/webapp/deploymentslot/DeploymentSlotNode.java
```diff
@@ -116,149 +123,42 @@ protected void refreshItems() {
...
-            return Node.getProgressMessage(AzureActionEnum.DELETE.getDoingName(), DeploymentSlotModule.MODULE_NAME, DeploymentSlotNode.this.name);
-        }
-
-        @Override
-        public TelemetryParameter getTelemetryParameter() {
-            return TelemetryParameter.WebApp.DeploymentSlot.DELETE;
-        }
+    @AzureOperation(value = "start deployment slot", type = AzureOperation.Type.ACTION)
+    private void start() {
+        AzureWebAppMvpModel.getInstance().startDeploymentSlot(subscriptionId, webAppId, slotName);
```
Flanker32, line 128: In function we get the status after start/stop/restart but deployment slot just render the action directly, shall we unify the behaviors?

shenqianjin: Will unity them including WebApp in a individual PR later.

Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/BasicActionBuilder.java
Flanker32: We may sync later for this parameter

Flanker32 (approved):

PR #4919, opened 2021-02-03
[telemetry 3]more AzureOperation on subscription/login related methods.

wangmingliang-ms:


Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/AuthMethodManager.java
```diff
@@ -165,6 +182,7 @@ public AuthMethodDetails getAuthMethodDetails() {
         return this.authMethodDetails;
     }
 
+    @AzureOperation(value = "update auth settings", type = AzureOperation.Type.TASK)
```
andxu, line 185: update-> persist

andxu (approved):

PR #4923, opened 2021-02-03
[telemetry 5]standalone op title bundle and move to azure-tools-core

wangmingliang-ms:
…e-tools-core.


Utils/azuretools-core/src/com/microsoft/azure/toolkit/lib/common/operation/AzureOperationBundle.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class AzureOperationBundle {
+
+    @NonNls
+    public static final String TITLES = "com.microsoft.azure.toolkit.operation.titles";
+    private static Provider provider;
+
+    public static synchronized void register(final Provider provider) {
+        if (AzureOperationBundle.provider == null) {
+            AzureOperationBundle.provider = provider;
```
andxu, line 41: should throw error when AzureOperationBundle.provider is non-null?

PR #4924, opened 2021-02-03
[telemetry 6] extract operation titles into message bundle with composite operation name as key

wangmingliang-ms:


Utils/azuretools-core/src/com/microsoft/azuretools/utils/WebAppUtils.java
```diff
@@ -466,7 +466,7 @@ public static String generateWebContainerPath(WebContainer webContainer) throws
     }
 
     @AzureOperation(
```
andxu, line 468: webapp|...?

andxu (approved):

PR #4960, opened 2021-02-23
Migrate Web App nodes in Azure explorer to app service library

Flanker32:
Migrate Web App nodes in Azure explorer to app service library


Flanker32:
Move to https://github.com/microsoft/azure-tools-for-java/pull/4972

PR #4991, opened 2021-03-05
Continue with warning when getting subscriptions & AADSTS50057 occurs on one tenant

maartenba:
Reported and seen in https://github.com/JetBrains/azure-tools-for-intellij/issues/438


andxu:
Thank you very much, @maartenba

PR #4998, opened 2021-03-09
Fix IDEA203 regression of Spark failure debug in local

wezhang:
The 203 API changes introduce the regression with local debugging Spark failure context with the following error message:

`java.nio.file.InvalidPathException: Illegal char <*> at index ...`


t-rufang (approved):
Thanks Wei!

PR #5030, opened 2021-03-18
Avoid reference AccessTokenAzureManager anywhere and doesn't stop whe…

andxu:
…n there are permission issue on some tenants

### DIFF
--- a/Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/srvpri/SrvPriManager.java
+++ b/Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/srvpri/SrvPriManager.java
@@ -13,6 +13,7 @@ import java.io.Writer;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.time.Duration;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Properties;
@@ -34,7 +35,10 @@ import com.microsoft.azuretools.authmanage.srvpri.step.RoleAssignmentStep;
 import com.microsoft.azuretools.authmanage.srvpri.step.ServicePrincipalStep;
 import com.microsoft.azuretools.authmanage.srvpri.step.Status;
 import com.microsoft.azuretools.authmanage.srvpri.step.StepManager;
-import com.microsoft.azuretools.sdkmanage.AccessTokenAzureManager;
+import com.microsoft.azuretools.sdkmanage.AzureManagerBase;
+
+import reactor.core.publisher.Mono;
+import reactor.util.retry.Retry;
 
 /**
  * Created by vlashch on 8/16/16.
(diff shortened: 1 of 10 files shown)


Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/srvpri/SrvPriManager.java
```diff
@@ -207,46 +208,39 @@ public void write(String str) throws IOException {
...
+    private static Mono<Boolean> checkArtifact(Reporter<String> fileReporter, Path filePath) {
         // here we try to use the file to check it's ok with retry logic
         fileReporter.report("Checking cred file...");
-        final int RETRY_QNTY = 5;
-        final int SLEEP_SEC = 10;
-        int retry_count = 0;
-        File authFiel = new File(filePath.toString());
-        while (retry_count < RETRY_QNTY) {
+
+        int maxRetry = 4;
```
Flanker32, line 215: May be a static constant

Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/srvpri/SrvPriManager.java
```diff
@@ -207,46 +208,39 @@ public void write(String str) throws IOException {
...
-                } else {
-                    fileReporter.report(String.format("Failed to check cred file after %s retries, error", retry_count, e.getMessage()));
+                    // make the retry work
                     throw e;
                 }
+                return false;
             }
-        }
+        }).retryWhen(Retry.backoff(maxRetry, Duration.ofSeconds(10))).onErrorResume(e -> {
+            fileReporter.report(String.format("Failed to check cred file after %s retries, error: %s", maxRetry, e.getMessage()));
```
Flanker32, line 239: Do we need to distinguish whether we meet no need to retry exception or we have retried for max time but no result?

Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/srvpri/SrvPriManager.java
```diff
@@ -212,47 +219,43 @@ public void write(String str) throws IOException {
...
+    // TODO: move this to common-lib
+    private static <T> Mono<T> retryOnError(int retryCount, Mono<T> observable, Function<Throwable, Boolean> shouldRetryOnError) {
+        return observable
+                .retryWhen(Retry.withThrowable(flux -> flux.zipWith(Flux.range(1, retryCount + 1), (throwable, count) -> {
+                    if (count < retryCount && (shouldRetryOnError.apply(throwable))) {
+                        return count;
+                    } else {
+                        throw Exceptions.propagate(throwable);
+                    }
+                }).flatMap(i -> Mono.delay(Duration.ofSeconds(((long) i * 10))))));
```
Flanker32, line 247: I think its good to wait for longer time when retry for a second time, but do we also need to reduce the retry count for create sp? Which used to be 40s but 150s now

Flanker32 (approved):

Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/srvpri/SrvPriManager.java
andxu: It is reasonable since 40s is not a precise time where normal role assignment will finish, we will not stick to this magic number

PR #5041, opened 2021-03-22
In CommandUtils.java, don't output error message when not needed

maartenba:
Make startup logs a bit cleaner by removing errors from the output when none are present.


andxu:
The target branch is develop, you can cherry pick it or we will handle it in another PR

maartenba:
Updated in https://github.com/microsoft/azure-tools-for-java/pull/5054, thanks @andxu !

PR #5058, opened 2021-03-25
apply new cache framework in webapp/functionapp listing

wangmingliang-ms:


Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/AuthMethodManager.java
```diff
@@ -148,6 +149,7 @@ public AzureManager getAzureManager() {
     }
 
     @AzureOperation(name = "account.sign_out", type = AzureOperation.Type.TASK)
+    @CacheEvict() // evict all caches on signing out
```
Flanker32, line 152: We may keep this as we do need to clean all azure related caches. but some cache is still valid after sign out, for instance, azure reference book meta data.

Utils/azuretools-core/src/com/microsoft/azuretools/core/mvp/model/function/AzureFunctionMvpModel.java
```diff
@@ -131,28 +126,29 @@ public void stopFunction(String sid, String appId) {
...
+            .flatMap((sd) -> listFunctionsInSubscription(sd.subscriptionId(), force).stream())
+            .collect(Collectors.toList());
+    }
+
+    @NotNull
+    @AzureOperation(
+        name = "function.list.java|subscription|selected",
+        type = AzureOperation.Type.SERVICE
+    )
+    @Preload
```
Flanker32, line 147: List Azure resources is only valid when user is signed in. however, users may not signed in preload time.(Even if user signed in before and restart IntelliJ, our auth manager may not be initialized). We may need to add condition for preload event and add more pre load triggers.

Utils/azuretools-core/src/com/microsoft/azuretools/core/mvp/model/webapp/AzureWebAppMvpModel.java
```diff
@@ -43,16 +43,12 @@
...
 
     private static final List<WebAppUtils.WebContainerMod> JAVA_8_JAR_CONTAINERS =
         Collections.singletonList(WebAppUtils.WebContainerMod.Java_SE_8);
     private static final List<WebAppUtils.WebContainerMod> JAVA_11_JAR_CONTAINERS = Collections.singletonList(
         WebAppUtils.WebContainerMod.Java_SE_11);
-
-    private AzureWebAppMvpModel() {
-        subscriptionIdToWebApps = new ConcurrentHashMap<>();
-    }
+    public static final String CACHE_SUBSCRIPTION_WEBAPPS = "subscription-webapps";
```
Flanker32, line 51: Do we also need to add cache for `getById` request?

Utils/azuretools-core/src/com/microsoft/azuretools/core/mvp/model/function/AzureFunctionMvpModel.java
wangmingliang-ms: > Even if user signed in before and restart IntelliJ, our auth manager may not be initialized
no, auth manager will read from file? anyway, it can work normally.

> We may need to add condition for preload event and add more pre load triggers.
as far as I know for this moment, we only need to re-trigger preloading when user sign in after startup/changed subscriptions.
we can complete to cover more cases step by step in the future.

Utils/azuretools-core/src/com/microsoft/azuretools/core/mvp/model/webapp/AzureWebAppMvpModel.java
wangmingliang-ms: which `getById` method?

Flanker32: method like `getWebAppById` to get specific resource

wangmingliang-ms: it's required if the request is slow, otherwise it's not necessary.

andxu (approved):

PR #5072, opened 2021-04-01
[Test] Revert "Fix linked non-cluster-default blob storage not working for HDI cluster issue"

t-rufang:
Reverts microsoft/azure-tools-for-java#4827


t-rufang:
This workaround is no longer needed because we have fixed the bug at #5073

Issue #5002, opened 2021-03-09
[IntelliJ] Failed to run Spark application with filled-in default Azure Blob storage account credential


t-rufang:
Repro steps:
1. Sign in Azure account
2. In Run/Configuration dialog, create a new HDInsight configuration. Select a Spark cluster with Azure blob storage account as default storage account.
3. [Key step] In `Job Upload Storage` part, select `Use Azure Blob to upload`. Fill in the storage account credential with **Cluster default storage account credential**. Save the config.
![image](https://user-images.githubusercontent.com/32627233/110479338-3d298580-8120-11eb-8970-b7551bc62fed.png)

4. Submit the Spark application. We will find the application will fail will following errors.
```
2021-04-02 05:55:46,248 INFO org.apache.livy.utils.LineBufferedStream: stdout: 21/04/02 05:55:46 INFO MetricsSystemImpl [main]: azure-file-system metrics system started
2021-04-02 05:55:46,273 INFO org.apache.livy.utils.LineBufferedStream: stdout: Exception in thread "main" org.apache.hadoop.fs.azure.AzureException: org.apache.hadoop.fs.azure.KeyProviderException: ExitCodeException exitCode=2: Error reading S/MIME message
2021-04-02 05:55:46,274 INFO org.apache.livy.utils.LineBufferedStream: stdout: 140390793635480:error:0D0680A8:asn1 encoding routines:ASN1_CHECK_TLEN:wrong tag:tasn_dec.c:1236:
2021-04-02 05:55:46,274 INFO org.apache.livy.utils.LineBufferedStream: stdout: 140390793635480:error:0D07803A:asn1 encoding routines:ASN1_ITEM_EX_D2I:nested asn1 error:tasn_dec.c:405:Type=CMS_ContentInfo
2021-04-02 05:55:46,274 INFO org.apache.livy.utils.LineBufferedStream: stdout: 140390793635480:error:0D0D106E:asn1 encoding routines:B64_READ_ASN1:decode error:asn_mime.c:192:
2021-04-02 05:55:46,274 INFO org.apache.livy.utils.LineBufferedStream: stdout: 140390793635480:error:0D0D40CB:asn1 encoding routines:SMIME_read_ASN1:asn1 parse error:asn_mime.c:517:
2021-04-02 05:55:46,274 INFO org.apache.livy.utils.LineBufferedStream: stdout: 
[...]

PR #5073, opened 2021-04-02
Fix job failure when using default blob storage account's credential for job upload storage

t-rufang:
- To fix issue #5002


wezhang (approved):
LGTM! Thank you Rui!

PR #5172, opened 2021-04-28
error occurs about resource connector after removing a runconfiguration and open configuration editor dialog again later

wangmingliang-ms:
similar to `singletonList`, `EmptyList` does not support `add`/`addAll`/... operations, which throws `UnsupportedOperationException` when called.

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/azure-intellij-resource-connector-lib/src/main/java/com/microsoft/azure/toolkit/intellij/connector/ConnectionRunnerForRunConfiguration.java
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/azure-intellij-resource-connector-lib/src/main/java/com/microsoft/azure/toolkit/intellij/connector/ConnectionRunnerForRunConfiguration.java
@@ -90,18 +90,14 @@ public class ConnectionRunnerForRunConfiguration extends BeforeRunTaskProvider<C
         public boolean isApplicableFor(@NotNull RunConfigurationBase<?> configuration) {
             final boolean applicable = configuration.getProject().getService(ConnectionManager.class)
                     .getConnections().stream().anyMatch(c -> c.isApplicableFor(configuration));
-            if (applicable) {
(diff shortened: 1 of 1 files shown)


andxu (approved):

PluginsAndFeatures/azure-toolkit-for-intellij/azure-intellij-resource-connector-lib/src/main/java/com/microsoft/azure/toolkit/intellij/connector/ConnectionRunnerForRunConfiguration.java
```diff
@@ -89,20 +90,26 @@ public boolean execute(@NotNull DataContext dataContext, @NotNull RunConfigurati
...
-                }
+            final List<BeforeRunTask<?>> tasks = configuration.getBeforeRunTasks();
+            final List<BeforeRunTask<?>> myTasks = tasks.stream().filter(t -> t instanceof MyBeforeRunTask).collect(Collectors.toList());
+            if (applicable && myTasks.isEmpty()) {
+                final MyBeforeRunTask task = new MyBeforeRunTask();
+                task.setEnabled(true);
+                this.addTask(configuration, task);
+            } else if (!myTasks.isEmpty()) {
+                tasks.removeAll(myTasks);
             }
```
shenqianjin, line 101: a little comments here? 
From above code lines, I don't know - why is it "addTask" when myTasks.isEmpty and "removeAll" when myTasks.isNotEmpty ?

wangmingliang-ms: it means there is a applicable resource connection for this run configuration when  `applicable==true`, which means we need to add a `MyBeforeRunTask` to this run configuration.
so when `applicable && myTasks.isEmpty`, we need to add a before run task.
but when **NOT** `applicable`, we need to remove the task.

PR #5176, opened 2021-04-28
Move AzureAnAction into the shared module azure-intellij-plugin-lib

jansorg:
This PR moves class `AzureAnAction` to allow reuse. The upcoming Azure AD submodule will make use of this.

- It adds a new dependency to the `azure-intellij-plugin-lib`, which is the same as in the main intellij module. It's needed because `AzureAnAction` uses its classes for telemetry.
- Please let me know if you'd like the declaration of this dependency in the main module instead of in the submodule

- CI is failing, including the unrelated `Azure-Toolkits-for-Java-CI-Eclipse`. I'm not sure what I can about this. I also can't view the CI logs, because this requires a login.


PluginsAndFeatures/azure-toolkit-for-intellij/azure-intellij-plugin-lib/build.gradle
```diff
@@ -1 +1,10 @@
+dependencies {
+    // as in the azure-toolkit-for-intellij module
+    compile 'com.microsoft.azuretools:azuretools-core:3.53.0-SNAPSHOT', {
```
wangmingliang-ms, line 3: as mentioned in other PRs, azuretools-core is to be discarded, so please don't add dependency on it.

wangmingliang-ms:
we are discarding this mechanism of telemetry. and `AzureAnAction` is to be discarded, so please use `AnAction` directly, and `AzureTelemeter`/`@AzureOperation`/`AzureTelemetry.@Properties`/`AzureTelemetry.@Property` for telemetering.

jansorg:
@wangmingliang-ms Thank you! In this case I'll close this PR as it's not required anymore.

PR #5177, opened 2021-04-29
Fix linked blob storage not working for livy interactive session issue

t-rufang:
#### Background
 - Currently when we start a livy interactive session,  some necessary config is missing if we use linked blob storage to store jar file. This PR is to add the missing config.
 
#### Todo
 - As you can see, the code I add is kind of duplicate at class `SparkBatchJobRunner`. Maybe we should refactor the code to put the code together at one place.


wezhang (approved):
LGTM! Thanks Rui!

PR #5226, opened 2021-05-15
Fix bug 1839028: Oauth report error when some tenants have permission issues

andxu:


Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/AuthMethodManager.java
```diff
@@ -83,7 +93,7 @@ private AuthMethodManager() {
             try {
                 initAuthMethodManagerFromSettings();
             } catch (Throwable ex) {
-                LOGGER.warning("Cannot restore login due to error: " + ex.getMessage());
+                LOGGER.warn("Cannot restore login due to error: " + ex.getMessage());
```
wangmingliang-ms, line 96: use AzureMessager if this message is to show to user.

wangmingliang-ms (approved):

Utils/azuretools-core/src/com/microsoft/azuretools/authmanage/AuthMethodManager.java
andxu: will refactor a lot of such cases in next sprint

PR #5293, opened 2021-05-25
SQL server and MySQL bug fix for end game

shenqianjin:


andxu (changes requested):
please split them in small dedicated PR

PR #5301, opened 2021-05-25
use getLabel instead of to string since region to string is not proper.

andxu:


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/forms/CreateVirtualNetworkForm.java
```diff
@@ -62,7 +62,7 @@ public void changedUpdate(DocumentEvent documentEvent) {
         subnetNameField.getDocument().addDocumentListener(docListener);
         subnetAddressRangeField.getDocument().addDocumentListener(docListener);
 
-        regionField.setText(region.toString());
+        regionField.setText(region.getLabel().toString());
```
Flanker32, line 65: no need to call toString here

andxu: will fix in new sprint

Flanker32 (approved):

PR #5303, opened 2021-05-25
Fix telemetry for web app deployment

Flanker32:
Fix telemetry for web app deployment


andxu (approved):
Please update toString method of model classes in libs.

PR #5320, opened 2021-06-01
[Warn Deprecated Azure Libs][#1848177, #1847666]get project libs and metadata of deprecated azure sdk libs.

wangmingliang-ms:


PluginsAndFeatures/azure-toolkit-for-intellij/azure-sdk-reference-book/src/main/java/com/microsoft/azure/toolkit/intellij/azuresdk/service/ProjectLibraryService.java
```diff
@@ -0,0 +1,46 @@
...
+        return libs.stream().filter(StringUtils::isNotBlank)
+                .map(PATTERN::matcher)
+                .filter(Matcher::matches)
+                .map(m -> new ProjectLibEntity(m.group(2).trim(), m.group(3).trim(), m.group(4).trim()))
+                .collect(Collectors.toList());
+    }
+
+    @Getter
+    @RequiredArgsConstructor
+    public static class ProjectLibEntity {
```
Flanker32, line 37: Do we need to track build tool in lib entity? (Though may not be used for now

PR #5346, opened 2021-06-08
Fix Narrator not reading browse text issue

t-rufang:
- We create this PR to fix [Accessibility bug 1738824](https://dev.azure.com/mseng/VSJava/_workitems/edit/1738824)


wezhang (approved):
LGTM! Thanks Rui!

PR #5351, opened 2021-06-09
[#1837563] Migrate to app service library to trigger functions

Flanker32:
Migrate to app service library to trigger functions

### DIFF
--- a/Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/function/FunctionNode.java
+++ b/Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/function/FunctionNode.java
@@ -23,15 +23,11 @@ import com.microsoft.tooling.msservices.serviceexplorer.Node;
 import com.microsoft.tooling.msservices.serviceexplorer.NodeActionEvent;
 import com.microsoft.tooling.msservices.serviceexplorer.NodeActionListener;
 import com.microsoft.tooling.msservices.serviceexplorer.WrappedTelemetryNodeActionListener;
+import lombok.RequiredArgsConstructor;
 import org.apache.commons.lang3.EnumUtils;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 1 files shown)


Flanker32:
AB#1837563

Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/function/FunctionNode.java
```diff
@@ -179,32 +134,17 @@ private String getFunctionHttpTriggerUrl() {
...
+        return String.format(HTTP_TRIGGER_URL_WITH_CODE, functionApp.hostName(), this.name, functionApp.getMasterKey());
     }
 
-    private Map<String, String> listFunctionKeys() {
-        // todo: imply with app service library
-        return Collections.EMPTY_MAP;
+    @RequiredArgsConstructor
+    static class TriggerRequest {
+        private final String input;
     }
```
shenqianjin, line 149: can we user only String to replace this class ?

Flanker32: trigger api in sdk will serialize input object into json map, so we could not pass the value string directly which will miss the key

Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/function/FunctionNode.java
```diff
@@ -96,52 +90,13 @@ private void trigger() {
                 triggerHttpTrigger(trigger);
                 break;
             case "timertrigger":
-                triggerTimerTrigger();
-                break;
-            case "eventhubtrigger":
-                triggerEventHubTrigger();
+                functionApp.triggerFunction(this.name, new Object()); // no input for timer trigger
```
Flanker32, line 93: try replace with null

andxu (approved):

PR #5426, opened 2021-06-23
Fix button tooltip not align with button text issue

t-rufang:
- We create this PR to fix accessibility bug [#1840628](https://dev.azure.com/mseng/VSJava/_workitems/edit/1840628)
 - After the fix, the UI looks like 
 
image


wezhang (approved):
LGTM! Thank you Rui!

PR #5427, opened 2021-06-23
[Bugfix 1854295] Local run will pop up error when when the IDE window exceeds 1

shenqianjin:


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/connector/database/DatabaseResourceConnection.java
```diff
@@ -116,19 +116,17 @@ private static Module getTargetModule(@NotNull RunConfiguration configuration) {
         return null;
     }
 
-    private Map<String, String> initEnv() {
+    private Map<String, String> initEnv(@Nonnull final Project project) {
         final Map<String, String> env = new HashMap<>();
-        final Module module = this.consumer.getModule();
```
andxu, line 121: shall we remove consumer?

shenqianjin: No. here aim to remove dependency on Intellij Module. it is different with our consumer - ModuleResource.

shenqianjin: our logic only depend on Project. it's unnecessary to add a "getModule" method on ModuleResource. and it may cause error in multi-projects case.

andxu (approved):

PR #5429, opened 2021-06-24
Fix refresh issue with cache for deployment slots and list functions

Flanker32:
Fix refresh issue with cache for deployment slots and list functions


shenqianjin (approved):
LGTM

PR #5437, opened 2021-06-25
Fix bug 1854234:  Edit json textbox service principal dialog will popup error

andxu:
[AB#1854234](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1854234)


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/intellij/ui/ServicePrincipalLoginDialog.java
```diff
@@ -256,33 +255,40 @@ public static String findTextInClipboard(Predicate<String> func) {
     }
 
     private void uiTextComponents2Json() {
-        if (intermediateState) {
+        if (!intermediateState.compareAndSet(false, true)) {
```
wangmingliang-ms, line 258: as mentioned in daily scrum.
you can solve the problem in a more elegant way by comparing the data-model(but not the view model) of the upper inputs-form and lower json-form, and then sync to the other when one is changed.

wangmingliang-ms (approved):

PR #5448, opened 2021-06-29
Fix name property be null when create sample scala project in Eclipse issue

t-rufang:
- We create this PR to fix the accessibility issue [#1840235](https://dev.azure.com/mseng/VSJava/_workitems/edit/1840235)


Flanker32 (approved):
LGTM~

PR #5508, opened 2021-07-12
fix miscellaneous bugs and performance issues about spring cloud properties/creation view

wangmingliang-ms:


PluginsAndFeatures/azure-toolkit-for-intellij/azure-intellij-plugin-lib/src/main/java/com/microsoft/azure/toolkit/intellij/common/AzureComboBox.java
```diff
@@ -158,6 +156,10 @@ private void refreshValue() {
                 super.setSelectedIndex(0);
             }
         } else {
+            final Object selected = this.getSelectedItem();
+            if (Objects.equals(selected, this.value) || (this.value instanceof ItemReference && ((ItemReference<?>) this.value).is(selected))) {
+                return;
```
Flanker32, line 161: Will it break the operation of set value with `ItemReference`? As line165 `items.stream().filter(i -> ((ItemReference<?>) this.value).is(i)).findFirst().ifPresent(this::setValue);` will call setValue with the item reference result, which may return here during setValue

Flanker32 (approved):
LGTM~

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/springcloud/creation/SpringCloudAppInfoAdvancedPanel.java
```diff
@@ -36,7 +37,7 @@ public SpringCloudAppInfoAdvancedPanel(@Nullable final SpringCloudCluster cluste
     }
 
     protected void onAppChanged(SpringCloudApp app) {
-        this.formConfig.updateForm(app);
+        AzureTaskManager.getInstance().runLater(() -> this.formConfig.updateForm(app));
```
Flanker32, line 40: Do we need to call runLater here? as it seems like an listener method

PR #5579, opened 2021-07-22
[Bug 1859080] Fix the bug of reporting error in azure explorer before sign in

andxu:


Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/mysql/MySQLModule.java
```diff
@@ -43,6 +43,10 @@ public void renderChildren(List<MySqlServer> servers) {
         }
     }
 
+    protected boolean requireSignedIn() {
+        return true;
+    }
+
```
shenqianjin, line 49: move this method to its abstract class and default as implementation ?

shenqianjin (dismissed):

Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/azure/AzureModule.java
```diff
@@ -136,6 +136,11 @@ public void setArcadiaModule(@NotNull HDInsightRootModule rootModule) {
         this.arcadiaModule = rootModule;
     }
 
+    @Override
+    protected void onNodeClick(NodeActionEvent e) {
+        super.onNodeClick(e);
+    }
+
```
wangmingliang-ms, line 143: remove it.

Utils/azure-explorer-common/src/com/microsoft/tooling/msservices/serviceexplorer/RefreshableNode.java
```diff
@@ -75,15 +77,15 @@ protected void onNodeClick(NodeActionEvent e) {
...
         if (!loading) {
             setLoading(true);
             try {
                 removeAllChildNodes();
-                if (forceRefresh) {
-                    refreshFromAzure();
+                if (AuthMethodManager.getInstance().isSignedIn() || this instanceof AzureModule) {
+                    refreshItems();
                 }
-                refreshItems();
```
wangmingliang-ms, line 86: these 2 methods is meaningful in old implementation:
`refreshFromAzure` requires reload the data from Azure(when user click the `fresh` button/menu item).
`refreshItems` indicates we can load from cache(when user click to expand a parent node.)

andxu: refreshFromAzure  are not used in current impl, remove it and will add it if we need it.

wangmingliang-ms (approved):

shenqianjin (approved):
LGTM

PR #5600, opened 2021-07-24
[Bugfix 1861059] Change with different subs, the username and database won't be clear while there is no server selected.

shenqianjin:
[AB#1861059](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1861059)


shenqianjin:
Replace with PR #5604

PR #5612, opened 2021-07-27
Fix bug: cannot restore login for non-global env

andxu:


Utils/azuretools-core/src/com/microsoft/azuretools/sdkmanage/IdentityAzureManager.java
```diff
@@ -132,13 +132,17 @@ public Subscription getSubscriptionById(String sid) {
...
+            Azure.az(AzureCloud.class).setByName(authMethodDetails.getAzureEnv());
+        }
         AuthType authType = authMethodDetails.getAuthType();
         try {
             if (authType == AuthType.SERVICE_PRINCIPAL) {
                 AuthConfiguration auth = new AuthConfiguration();
                 auth.setType(AuthType.SERVICE_PRINCIPAL);
                 auth.setClient(authMethodDetails.getClientId());
                 auth.setTenant(authMethodDetails.getTenantId());
+                auth.setEnvironment(Azure.az(AzureCloud.class).get());
```
Flanker32, line 145: Do we need to set env here? Can we could keep it empty and make `AzureAccount` get value from `Azure.az(AzureCloud.class)`

Flanker32 (approved):

Utils/azuretools-core/src/com/microsoft/azuretools/sdkmanage/IdentityAzureManager.java
andxu: It is a design: AzureAccount and AzureCloud are different concepts, env in AzureAccount is the env of login result, env in AzureCloud is the env of next login when not logged in and when it is logged in the env in AzureCloud is not changed, in this place it is to login, so we need to set AzureCloud. But due to the refactor of AUTH Configuration is not done, we need to both call auth.setEnvironment and set AzureCloud.

PR #5615, opened 2021-07-27
Fix npe issue for create new docker deployment configuration

Flanker32:
Fix npe issue for create new docker deployment configuration


andxu (approved):
LGMT

PR #5619, opened 2021-07-27
Fix default region for app service deployment

Flanker32:
Get app service default region by azure environment, fix wrong default region in Azure China.

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/lib/appservice/AppServiceConfig.java
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/lib/appservice/AppServiceConfig.java
@@ -5,12 +5,12 @@
 
 package com.microsoft.azure.toolkit.lib.appservice;
 
+import com.azure.core.management.AzureEnvironment;
 import com.microsoft.azure.toolkit.lib.Azure;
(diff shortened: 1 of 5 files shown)


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/lib/appservice/AppServiceConfig.java
```diff
@@ -50,4 +57,16 @@
         result.put("pricingTier", Optional.ofNullable(pricingTier).map(PricingTier::getSize).orElse(StringUtils.EMPTY));
         return result;
     }
+
+    public static Region getDefaultRegion() {
+        final String environment = AzureEnvironmentUtils.azureEnvironmentToString(Azure.az(AzureAccount.class).account().getEnvironment());
```
andxu, line 62: bad code to switch on azureEnvironmentToString(which is the display ui logic), use if (Azure.az(AzureAccount.class).account().getEnvironment() == AzureEnvironment.CHINA ?

andxu: we have a list-region api, shall we use the first region of that list?

andxu (approved):
LGTM

PR #5642, opened 2021-08-03
Fix app service creation dialog takes much time to initialize

Flanker32:
### Issue
App service creation dialog will froze for few seconds after pop up. The reason is we call force refresh method in `setSubscription()` of `ServicePlanComboBox`，however, this method will be called in EDT during dialog initialization, which may froze the UI

### Solution
Do force refresh action in `loadItems` which will be called in background thread

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/appservice/serviceplan/ServicePlanComboBox.java
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/appservice/serviceplan/ServicePlanComboBox.java
@@ -16,6 +16,7 @@ import com.microsoft.azure.toolkit.lib.appservice.model.PricingTier;
 import com.microsoft.azure.toolkit.intellij.common.AzureComboBox;
 import com.microsoft.azure.toolkit.intellij.common.Draft;
 import com.microsoft.azure.toolkit.lib.appservice.DraftServicePlan;
+import com.microsoft.azure.toolkit.lib.common.cache.CacheManager;
 import com.microsoft.azure.toolkit.lib.common.model.Region;
 import com.microsoft.azure.toolkit.lib.common.model.Subscription;
 import com.microsoft.azure.toolkit.lib.common.operation.AzureOperation;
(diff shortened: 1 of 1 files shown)


PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/appservice/serviceplan/ServicePlanComboBox.java
```diff
@@ -55,8 +55,6 @@ public void setSubscription(Subscription subscription) {
             this.clear();
             return;
         }
-        // force refresh service plan when switch subscription
-        Azure.az(AzureAppService.class).subscription(subscription.getId()).appServicePlans(true);
         this.refreshItems();
```
wangmingliang-ms, line 58: you may just wrap these 2 lines with `runOnPooledThread`

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/appservice/serviceplan/ServicePlanComboBox.java
```diff
@@ -114,7 +112,7 @@ protected String getItemText(final Object item) {
                     .collect(Collectors.toList()));
             }
             final List<AppServicePlanEntity> remotePlans = Azure.az(AzureAppService.class)
-                .subscription(subscription.getId()).appServicePlans().stream().map(IAppServicePlan::entity)
+                .subscription(subscription.getId()).appServicePlans(true).stream().map(IAppServicePlan::entity)
```
wangmingliang-ms, line 115: this will cause every loading skips cache, is this expected?

PluginsAndFeatures/azure-toolkit-for-intellij/src/com/microsoft/azure/toolkit/intellij/appservice/serviceplan/ServicePlanComboBox.java
Flanker32: Changed to clear service plan cache instead of always force refresh.

For force refresh issue, now we usually add event listener to force refresh the explorer after resource creation(webapp/function), however, we didn't apply same framework to service plan, and I think we may need to do the cache refresh job in library rather than the toolkit

PR #5607, opened 2021-07-26
[Bug fix] add a new doSignIn method to replace the old one which has bug on EDT

andxu:
image
image

PR #5643, opened 2021-08-03
[Bugfix 1863833]: support cancel login

andxu:
## What problem is this PR solving?
In this PR https://github.com/microsoft/azure-tools-for-java/pull/5607,  I have missed one statement of deviceLoginUI.setFuture during the move from SignInWindow to SignInAction. And even in previous implementation, it is non-cancelable for getting device code, so I add the logic of getting device code into checkCancelled. 

## What has been done in this PR?
 1. originally checkCancelled only returns AuthMethodDetail, in this PR, it is generalized with T, which may return AuthMethodDetail or DeviceCodeAccount, in future it is the key api for support cancelable job and move to common lib

[AB#1863833](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1863833),


Flanker32 (approved):

shenqianjin (commented):
LGTM

PR #5655, opened 2021-08-04
Fix validation error accessibility issue of linking a cluster

wezhang:
Refer to the screen recording:
LinkAClusterErrorAccessibilityIssue


t-rufang (approved):
Smart solution! Thank you!

PR #5658, opened 2021-08-06
Fix function creation issue with multi subscription

Flanker32:
### Issue to resolve
Function deployment will failed when multi subscription was selected in IntelliJ, as we didn't pass target subscription to `Azure.az(AzureAppService.class)`, which will use default selected subscription and will throw exception when multi subscriptions are selected

image

### Solution
always pass subscription to `AzureService`


andxu (approved):
LGTM, but still need refactor the logic of creating rg in task

PR #5659, opened 2021-08-06
follow the gradle project convention for source set.

andxu:
Current layout disables the cache and makes compileJava goal very slow, with the change in this PR, the compilation speed is very fast.
image
The convention is at: https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_project_layout


wezhang (approved):
LGTM!

PR #5666, opened 2021-08-09
rename IAzureEntityManager -> IAzureResource

wangmingliang-ms:


shenqianjin:
why so many commits in the PR?

PR #5678, opened 2021-08-11
[#1867031] set TemplatePresentation so that actions are accessible from action palatte.

wangmingliang-ms:
[AB#1867031](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1867031)


shenqianjin (approved):
LGTM

PR #5683, opened 2021-08-11
add registerHandler to AzureActionManager and the latter registered handlers has higher priority

wangmingliang-ms:


andxu (approved):
LGTM

PR #5688, opened 2021-08-11
[Intellij Accessibility]NVDA does not announce webhdfs labels

yanjungao718:
add announce for webhdfs panel labels


t-rufang (approved):
Thanks, Yanjun! LGTM

PR #5693, opened 2021-08-11
[Bug 1853307]Fix label of combobox not read by NVDA issue

t-rufang:
- We create this PR to fix part of the bug [#1853307](https://dev.azure.com/mseng/VSJava/_workitems/edit/1853307)


Flanker32 (approved):
LGTM~

PR #5704, opened 2021-08-16
Suppress fake credential warning for test cases

Flanker32:
### Issues to resolve:
Credential Scanner found some credential issues, which are caused by fake credentials in test cases

### Solution
Suppress fake credentials in resource file


andxu (approved):
LGTM

PR #5706, opened 2021-08-16
Add application insights dependency to Utils

Flanker32:
### Issues to resolve:
Add application insights dependency to Utils, as we have removed the dependency from toolkits-lib.
Refers https://github.com/microsoft/azure-maven-plugins/pull/1701

Todo: migrate application insights to Track2 SDK, remove track one dependency


andxu (approved):
LGTM

PR #5739, opened 2021-08-23
Sync function before list function triggers

Flanker32:
### Issues to resolve:
Function triggers will take quite a bit time to refresh (30~60s) after deployment
- For new created triggers, triggers could not be listed after deployment
- For update, old triggers will still shown in portal/deployment status

### Solution
- Sync function before list function triggers, refers https://github.com/Azure/azure-functions-core-tools/blob/3.0.3568/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs#L452


andxu (approved):
LGTM

PR #5746, opened 2021-08-25
Update plugin.xml path in intellij version helper

Flanker32:
Update plugin.xml path in intellij version helper, which may make release bits lack ide version


andxu (approved):
LGTM

PR #5747, opened 2021-08-25
Update intellij resource path in configuration

Flanker32:
Update intellij resource path in configuration as we have move it from `/resources` to `/src/main/resources`


andxu (approved):
LGMT

wezhang (approved):
LGTM! Thanks @Flanker32 ！

PR #5774, opened 2021-09-06
refactor to make AzureResource creatable from AzureService node.

wangmingliang-ms:


Utils/azure-toolkit-ide-libs/azure-toolkit-ide-common-lib/src/main/java/com/microsoft/azure/toolkit/ide/common/action/ResourceCommonActionsContributor.java
```diff
@@ -92,7 +92,16 @@ public void registerActions(AzureActionManager am) {
...
+                .title(s -> Optional.ofNullable(s).map(r -> {
+                    String name = r.getClass().getSimpleName();
+                    if (r instanceof IAzureResource) {
+                        name = ((IAzureResource<?>) r).name();
+                    } else if (r instanceof AzureService) {
+                        name = ((AzureService) r).name();
+                    }
+                    return title("common|resource.create", name);
+                }).orElse(null))
+                .enabled(s -> s instanceof IAzureResource || s instanceof AzureService);
```
Flanker32, line 104: Can actions be accessed in IntelliJ action palette for service without create action

wangmingliang-ms: yes, this is a problem but not serious, I would fix this problem in later PRs. 
**Actions should be disabled if there are not suitable action handler.**

Flanker32 (approved):

PR #5794, opened 2021-09-12
Add short cut for refresh action in select subscription dialog

Flanker32:
Add short cut for refresh action in select subscription dialog, for issue https://dev.azure.com/mseng/VSJava/_workitems/edit/1853397


andxu (approved):
LGTM

PR #5807, opened 2021-09-14
[Task 1870806] Migrate to IAzureBaseResource in ide action manager

Flanker32:
- Migrate to IAzureBaseResource in ide action manager
- Add enable condition for default handler of common action start/restart/stop
- Disable action if no handler is enabled in IntelliJ


andxu (approved):
LGTM

PR #5833, opened 2021-09-18
2. add dataId to resource and allow creating resource from data id for azure resources.

wangmingliang-ms:
…zure resources.


andxu (approved):
LGTM

PR #5844, opened 2021-09-23
Add inbound port panel to basic panel of vm creation dialog

Flanker32:
Add inbound port panel to basic panel of vm creation dialog and keep consistent with network configuration

image


andxu:
remove the none allow selected line

andxu (approved):
LGTM

PR #5854, opened 2021-09-24
Fix possible npe for action with authentication required

Flanker32:
Fix possible npe for action with authentication required


andxu (approved):
LGTM

PR #5857, opened 2021-09-26
Fix fail to load plugin error

### LINKED ISSUE #5825: InteliJ IDEA plugin: Empty value for config-file causes issues on plugin load
Reported here https://stackoverflow.com/q/69203962/2000323

I've [checked](https://github.com/microsoft/azure-tools-for-java/blob/develop/PluginsAndFeatures/azure-toolkit-for-intellij/src/main/resources/META-INF/plugin.xml#L52) and the plugin has empty value for `config-file` attribute in the plugin descriptor:

```
  <depends optional="true" config-file="">org.intellij.scala</depends>
```

Now it may lead to problem when loading the plugin, see also this comment: https://youtrack.jetbrains.com/issue/IDEA-270963#focus=Comments-27-5056928.0-0

t-rufang:
- To fix #5825 
 - Refer to solution in https://youtrack.jetbrains.com/issue/IDEA-270963#focus=Comments-27-5056928.0-0

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/resources/META-INF/withScala.xml
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/resources/META-INF/withScala.xml
@@ -1,2 +1,0 @@
-<idea-plugin>
-</idea-plugin>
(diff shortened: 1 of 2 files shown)


Flanker32 (approved):
LGTM~

PR #5866, opened 2021-09-28
Bump toolkitlib version to 0.12.1

andxu:


Flanker32 (approved):
LGTM

Utils/pom.xml
```diff
@@ -317,7 +317,7 @@
             <dependency>
                 <groupId>org.apache.httpcomponents</groupId>
                 <artifactId>httpclient</artifactId>
-                <version>${httpclient.version}</version>
+                <version>4.5.2</version>
```
Flanker32, line 320: may keep origin parameter reference but change the value

PR #5868, opened 2021-09-28
[#1881316] Right click a running spring cloud app, only start is enabled, stop and restart is disabled

wangmingliang-ms:
[AB#1881316](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1881316)


Flanker32 (approved):
LGTM

PR #5869, opened 2021-09-28
Fix invalid cache name for app service plan

Flanker32:
Fix invalid cache name for app service plan, which may make service plan always use cached list


andxu (approved):
LGTM

PR #5924, opened 2021-10-13
#1885354: configuration of storage blob/file endpoint is not correct

wangmingliang-ms:
[AB#1885354](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1885354)


Flanker32 (approved):
LGTM

PR #5954, opened 2021-10-20
[Feature #1880915] generalize common components and implement their eclipse version, so that we can even share UI code between IJ and eclipse

wangmingliang-ms:
[AB#1880915](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1880915)


andxu (dismissed):
LGTM

Flanker32 (approved):
LGTM

PR #5977, opened 2021-10-22
Fix issues in eclipse web app explorer

Flanker32:
- Fix package name and download action for eclipse action contributor
- Fix name prefix for explorer icons
- Clear Azure tree nodes after subscription changes
- Fix dialog dispose issue in sign in window


wangmingliang-ms (approved):
LGTM

PR #5981, opened 2021-10-22
Fix issues in app service explorer

Flanker32:
- Update icon path for file nodes in azure explorer
- Fix AzureTreeNode was not removed after sign out
- Fix UI disposed issue in azure sign in dialog


wangmingliang-ms (approved):
LGTM

PR #5990, opened 2021-10-25
Fix shotcut conflicts in the customized UI component

t-rufang:
- We create this PR to fix [#1853291](https://dev.azure.com/mseng/_workitems/edit/1853291). With this PR there will be no shortcut conflicts for the `browse` button and the `expand` button
1853291_keystroke


wezhang (approved):
LGTM! Thanks Rui!

PR #5995, opened 2021-10-26
Fix issues for eclipse appservice

Flanker32:
- Fix NPE for unknown file extension in azure explorer
- Fix empty slot name while create deployment slot


wangmingliang-ms (approved):
LGTM

PR #5996, opened 2021-10-26
Fix config not applied when starting livy interactive console issue

t-rufang:
- We create this PR to fix bug [#1867382](https://dev.azure.com/mseng/VSJava/_workitems/edit/1867382)


wezhang (approved):
LGTM!

PR #6001, opened 2021-10-26
Add theme support to HDI job view

wezhang:
Also add env HDI_JOBVIEW_ROOTDIR for easy development:

image


t-rufang (approved):
Thanks Wei!

PR #6008, opened 2021-10-27
[Test #1886478] public ip is required in vm creation

Flanker32:
AB##1886478 Fix public ip is required in vm creation


andxu (approved):
LGTM

PR #6035, opened 2021-11-03
only ApplicationSettings of one connection is successfully set even if there are multiple connections

wangmingliang-ms:


Flanker32 (approved):
LGTM

PR #6044, opened 2021-11-05
Fix table not focusable accessibility issue

t-rufang:
- We create this PR to fix the table not focusable accessibility issue [#1870841](https://dev.azure.com/mseng/VSJava/_workitems/edit/1870841)
1870841_table_focus

## References
1. Data Grid: https://www.w3.org/TR/wai-aria-practices/examples/grid/dataGrids.html#


wezhang (approved):
LGTM! Thanks Rui!

PR #6072, opened 2021-11-10
[Task #1891083] Package dependencies into fat jar to meet `osgi.bundlefile.limit`

Flanker32:
### Issues to resolve
This PR is to resolve the perf issue caused by osgi limit for Azure Toolkit for Eclipse 
[`osgi.bundlefile.limit`](https://www.eclipse.org/virgo/documentation/virgo-documentation-3.6.4.RELEASE/docs/virgo-user-guide/html/ch13.html) restrict the number of jar files the framework will keep open, which defaults to be `100`. However, toolkit for eclipse introduced more dependency jars(250+) than the threshold values, which makes Eclipse platform need to reload jars in class path occasionally and affected the perf

### What this PR do
- Package the dependencies for Azure toolkit with maven-shade-plugin to reduce the jar file in class path
- Remove unnecessary artifact dependencies from azuretools.core

[AB#1891083](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1891083)


wangmingliang-ms (approved):
LGTM

PR #6086, opened 2021-11-15
#1894063: UI blocks if user can not access github when open sdk reference book

wangmingliang-ms:
[AB#1894063](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1894063): UI blocks if user can not access github when open sdk reference book


Flanker32 (approved):
LGTM

PR #6161, opened 2021-12-03
Add job view focus loop

wezhang:
To enable job view focus loop in elements on page when press 'tab'
JobviewFocusLoopAccessibilityFix


t-rufang (approved):
Thanks, Wei! It would be better to provide a gif to indicate the changes.

wezhang:
> Thanks, Wei! It would be better to provide a gif to indicate the changes.

Np, Provided.

PR #6163, opened 2021-12-05
Migrate to app service implementation instead of interface

Flanker32:
Migrate to app service implementation instead of interface


wangmingliang-ms (approved):
LGTM

PR #6172, opened 2021-12-07
revert 7f0f9991

wangmingliang-ms:
…ception: Could not find org.codehaus.jackson:jackson-core-asl:1.9.13-atlassian-6.

PR #6174, opened 2021-12-07
fix cve issues except jackson related

yanjungao718:
this pr is based on https://github.com/microsoft/azure-tools-for-java/pull/6172, re-commit changes except jackons related after reverting


wangmingliang-ms (approved):
LGTM, thanks!

PR #6183, opened 2021-12-09
#1881356: [Test] No default value for CPU and Memory if click more settings quickly

wangmingliang-ms:
[AB#1881356](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1881356): [Test] No default value for CPU and Memory if click more settings quickly


andxu (approved):
LGTM

PR #6213, opened 2021-12-15
some exceptions are not caught by our exception handler but caught by IntelliJ when Mono is used.

wangmingliang-ms:


andxu (approved):
LGTM

PR #6224, opened 2021-12-16
FEATURE#1899919: add Azure's account registration link in login dialog, so that user can register account conveniently

wangmingliang-ms:
[AB#1899919](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1899919): add Azure's account registration link in login dialog, so that user can register account conveniently


andxu (approved):
LGTM

PR #6233, opened 2021-12-16
#1886504: [Test] Delete and open in portal icons are different between resources

wangmingliang-ms:
[AB#1886504](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1886504): [Test] Delete and open in portal icons are different between resources


andxu (approved):
LGTM

PR #6248, opened 2021-12-20
Extract config model for function/webapp to share within Eclipse and IntelliJ

Flanker32:
## What this PR do
IntelliJ keeps its own app service and related models like `Draft` , this pr move the model to ide library so that Eclipse and IntelliJ could share same model and related convert logic.


andxu (approved):
LGTM

PR #6251, opened 2021-12-21
Fix bug: 1. create function wizard and then finish at first step, wil…

andxu:
…l generate empty artifact id(#1903982)

2. Fix bug java.lang.NoClassDefFoundError: org/codehaus/plexus/components/io/attributes/Java7Reflector(#1904016)
3. Fix bug #1903998 An error pops up while adding new function class "RabbitMQTrigger"


Flanker32 (approved):
LGTM

PR #6270, opened 2021-12-27
#1904008: [Test]Create function class failed on Mac due to unhandled event loop exception

wangmingliang-ms:
[AB#1904008](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1904008) : [Test]Create function class failed on Mac due to unhandled event loop exception


Flanker32 (approved):
LGTM

PR #6342, opened 2022-01-21
[Task #1872026] Remove reference for deprecated entity() for app service

Flanker32:
Remove reference for deprecated entity() for app service, refers https://github.com/microsoft/azure-maven-plugins/pull/1914


wangmingliang-ms (approved):
LGTM

PR #6390, opened 2022-02-11
[#1917246] add “Create” action (link to portal) on `Spring Cloud` node, so that user can jump directly to spring cloud service creation page on portal.

wangmingliang-ms:
[AB#1917246](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1917246): add “Create” action (link to portal) on `Spring Cloud` node, so that user can jump directly to spring cloud service creation page on portal.


Flanker32 (approved):
LGTM

PR #6448, opened 2022-02-28
Fix dependencies for gradle function project

### LINKED ISSUE #6376: Intellij - For gradle projects, azure function debug does not pre-package inter-module  dependencies, resulting in classdefnotfound error
For gradle projects, create a vanilla gradle project
then create two modules:
1. with an azure function using v4
2.  some other gradle module with a custom class

3. In the build.gradle of the azure function module refer to the module from step two in the dependencies block
4. In the function refer to the custom class from the module created in step 2.

Debug the function using the azure functions debug run configuration template, when it hits the custom class it will result in a "classdefnotfound" exception

Debugging the function using the azure functions gradle plugin + remote java debug intellij run configuration results in the function proceeding as expected

Flanker32:
We used to use `ExternalProjectDataCache.getRootExternalProject` to get gradle function project module and get its dependencies for local run and deploy, however, this did not works for function in gradle inter-module. To solve this issue, this PR will list sub modules of current project and get correspond `ExternalProject` for target inter-moudle, this PR also fix the method to get artifact for gradle project dependency.

fixes #6376

N/A

Any other comments?
-------------------
N/A

Has this been tested?
---------------------------
- [x] Tested


wangmingliang-ms (approved):
LGTM

Issue #1991, opened 2018-09-12
[IntelliJ] "could not find the local jar package for Artifact" still displayed


tankungx:
Build: PR1945 windows

Repro Steps:
1. Right click a cluster, choose "Submit Job", wait all null information disappear.

2. Choose "Artifact from local disk", and browse jar file.

Actual:
still display "could not find the local jar package for Artifact"
![image](https://user-images.githubusercontent.com/30617450/45416124-c7a8df80-b6b1-11e8-8ca2-0161b7f041be.png)

tankungx:
add lables

v-jiche:
Verified. Fixed
azure-toolkit-for-intellij-2018.1.PR.2070.09-18-2018 (1)

PR #6546, opened 2022-04-12
[Do Not Merge] Fix URIBuilder behavior change after upgrading httpclient: 4.5.2 -> 4.5.13

t-rufang:
Please don't merge this PR unless [#1991 in repo azure-maven-plugins](https://github.com/microsoft/azure-maven-plugins/pull/1991) is merged first.

In order to fix CVE issue [WS-2017-3734](https://dev.azure.com/mseng/VSJava/_componentGovernance/445/alert/12284?typeId=119493), we need to upgrade dependency `org.apache.httpcomponents:httpclient ` from  version `4.5.2` to latest version `4.5.13`. But if we do the upgrade in [Repo azure-maven-plugin](https://github.com/microsoft/azure-maven-plugins/pull/1991) , UT will fail. After debugging, we found the UT failure is caused by behavior  change of `URIBuilder`.

If we pass `"/subPath"` to the following code snippet, `httpclient:4.5.2` will return `/subPath` but `httpclient:4.5.13` will return `//subPath`.

```Java
new URIBuilder().setPath(rawPath).build()
```
We create this PR to fix such behavior change.

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [ ] Tested

### DIFF
--- a/Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/AzureStorageUri.java
+++ b/Utils/hdinsight-node-common/src/com/microsoft/azure/hdinsight/common/AzureStorageUri.java
@@ -135,7 +135,7 @@ abstract public class AzureStorageUri {
      */
     public static String encodeAndNormalizePath(String rawPath) {
         try {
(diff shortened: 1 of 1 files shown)


yanjungao718 (approved):
LGTM

Issue #6696, opened 2022-06-09
Unable to run debug on azure java function on intellij (2022.1) with azure tookit (3.65.1)


UQsolutions:
Installed Azure Toolkit (3.65.1-2022.1) on intellij Community version 2.22.1 (Build #IC-221.5787.30, built on June 1, 2022)
I created the default HttpTrigger function.

When i simply run the function, it is successful and this is the output:
Extension bundle specified, skip install extension
Using port : 7,071

Azure Functions Core Tools
Core Tools Version:       4.0.4544 Commit hash: N/A  (64-bit)
Function Runtime Version: 4.3.2.18186

Functions:

	HttpExample: [GET,POST] http://localhost:7071/api/HttpExample

For detailed output, run func with --verbose flag.
[2022-06-09T13:24:43.891Z] Worker process started and initialized.
[2022-06-09T13:24:43.907Z] [WARNING] {JavaFunctionBroker.addJavaAnnotationLibrary}: Customer is not providing java annotation files, they may not use the latest version of java library and maven plugin.
[2022-06-09T13:24:48.255Z] Host lock lease acquired by instance ID '0000000000000000000000008AD5A5E9'.

But have been unable to start debugging.     When i click debug, i get this in the Run Functions console:
Extension bundle specified, skip install extension
Using port : 7,071

Azure Functions Core Tools
Core Tools Version:       4.0.4544 Commit hash: N/A  (64-bit)
Function Runtime Version: 4.3.2.18186

[2022-06-09T13:25:54.134Z] Listening for transport dt_socket at address: 7071
[2022-06-09T13:25:54.234Z] Failed to start a new language worker for runtime: java.
[2022-06-09T13:25:54.234Z] System.Private.CoreLib: A task was canceled.

I also get a popup notifications:
Error running 'azure functions': Unable to open debugger port (localhost:7071): java.io.IOException "connection is closed
Failed to run function CLI command from staging folder...

I removed all settings, cache and cleared the intellij installation and re installed, but still the problem persists.

I think it might be trying to use the Azure function end point port 7071 for the debug as well.

Not sure where this is happening.

Steps to reproduce:
With the version of Intellij mentioned, create a default HttpTrigger azure function and run debug.

UQsolutions:
[...]

PR #6717, opened 2022-06-20
Fix function local debug port may be same as function host

Flanker32:
Skip function host port when find free port for debug session, fix function local debug port may be same as function host

#6696

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [ ] Tested


wangmingliang-ms (approved):
LGTM

PR #6793, opened 2022-07-25
Fix issues in getting start experience

Flanker32:
- Fix signed in account's subscription is not validated in sign in task
- Do not break execution but show as `partial succeed` for unnecessary steps in phase
- Show warning instead of throw exception when failed to create application insights during create function app
- Chande default region to create app service to source central us

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [ ] Tested


wangmingliang-ms (approved):
LGTM

PR #7401, opened 2023-03-20
Azure Monitor: Extend date range filter with time units

maksyml85:
Adding time units (e.g. hours/minutes/seconds) to a time range component of Azure Monitor will allow to fetch log entries (especially historical ones) more precisely when there is only 5000 of maximal records we can fetch according to settings. 

Seems no

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [x] Tested


wangmingliang-ms (approved):
LGTM

maksyml85:
> @microsoft-github-policy-service agree

ivywei0125:
Hi @maksyml85, thanks for the contribution.  Do you want to expand the range of time to support **custom setting** hours, minutes and seconds?
The DateFilterComponent only support date picker, it doesn't support modifying time units even though pass with HH:mm:ss. 
We would implement the feature of setting hours, minutes and seconds as soon as possible. Please feel free contact us if you have any other questions/features.

maksyml85:
Hi @ivywei0125 . Thanks for your reply. Exactly, I want to expand time range and so have ability to fetch for more specific periods of time.
Yes, formally I've just put date in a date time format chosen from datepicker into text field with ability to modify in place, and it looks ugly in a bit. That's great if you are going to improve that. Could you please provide with a link to a related issue?

ivywei0125:
Related issue:
https://github.com/microsoft/azure-tools-for-java/issues/7414

Issue #10788, opened 2025-07-04
Azure Toolkit For IntelliJ not working


MSP8649:
After installing the **Azure Toolkit For IntelliJ** plugin, have restarted the IDE. Plugin is installed successfully but could see below error.

```
java.lang.NullPointerException: Cannot invoke "com.microsoft.azure.toolkit.intellij.java.sdk.utils.RuleConfigLoader.getRuleConfigs()" because the return value of "com.microsoft.azure.toolkit.intellij.java.sdk.utils.RuleConfigLoader.getInstance()" is null
	at com.microsoft.azure.toolkit.intellij.java.sdk.analyzer.DisableAutoCompleteCheck.buildVisitor(DisableAutoCompleteCheck.java:40)
	at com.intellij.codeInspection.LocalInspectionTool.buildVisitor(LocalInspectionTool.java:128)
	at com.intellij.codeInspection.InspectionEngine.createVisitor(InspectionEngine.java:55)
	at com.intellij.codeInsight.daemon.impl.InspectionRunner.lambda$inspect$11(InspectionRunner.java:155)
	at com.intellij.codeInspection.InspectionEngine.withSession(InspectionEngine.java:240)
	at com.intellij.codeInsight.daemon.impl.InspectionRunner.inspect(InspectionRunner.java:147)
	at com.intellij.codeInsight.daemon.impl.LocalInspectionsPass.lambda$collectInformationWithProgress$4(LocalInspectionsPass.java:149)
	at com.intellij.codeInsight.daemon.impl.HighlightInfoUpdaterImpl.lambda$runWithInvalidPsiRecycler$42(HighlightInfoUpdaterImpl.java:869)
	at com.intellij.codeInsight.daemon.impl.ManagedHighlighterRecycler.runWithRecycler(ManagedHighlighterRecycler.java:87)
	at com.intellij.codeInsight.daemon.impl.HighlightInfoUpdaterImpl.runWithInvalidPsiRecycler(HighlightInfoUpdaterImpl.java:850)
	at com.intellij.codeInsight.daemon.impl.LocalInspectionsPass.collectInformationWithProgress(LocalInspectionsPass.java:161)
	at com.intellij.codeInsight.daemon.impl.ProgressableTextEditorHighlightingPass.doCollectInformation(ProgressableTextEditorHighlightingPass.java:88)
	at com.intellij.codeHighlighting.TextEditorHighlightingPass.collectInformation(TextEditorHighlightingPass.java:70)
[...]

PR #11254, opened 2025-09-17
Use singleton pattern for rule loading to avoid NPE

samvaity:
Avoids NPE during early phases or race conditions in rule loading, adds a null-check RuleConfigLoader.getInstance() and fall back to an empty rules map

Related https://github.com/microsoft/azure-tools-for-java/issues/10788

Has this been tested?
---------------------------
- [X] Tested


PluginsAndFeatures/azure-toolkit-for-intellij/azure-intellij-plugin-java-sdk/src/main/java/com/microsoft/azure/toolkit/intellij/java/sdk/analyzer/ConnectionStringCheck.java
```diff
@@ -23,7 +25,9 @@ public class ConnectionStringCheck extends LocalInspectionTool {
 
     @Override
     public @NotNull PsiElementVisitor buildVisitor(@NotNull ProblemsHolder holder, boolean isOnTheFly) {
-        return new ConnectionStringCheckVisitor(holder, RuleConfigLoader.getInstance().getRuleConfigs());
+        final RuleConfigLoader loader = RuleConfigLoader.getInstance();
+        final Map<String, RuleConfig> rules = (loader != null) ? loader.getRuleConfigs() : Collections.emptyMap();
```
srnagar: Can we ensure that `loader` is never `null`? RuleConfigLoader.getInstance() should return a singleton and looking at the RuleConfigLoader, the configuration is loaded lazily which is causing this issue. Can we eagerly load the config and ensure getInstance always returns a valid instance?

srnagar (approved):

wangmingliang-ms (approved):
LGTM

PR #11939, opened 2026-04-09
Enable Azure Skills for GitHub Copilot

srnagar:
This pull request introduces a new post-startup activity for automatically installing or updating Azure Skills in the Azure MCP IntelliJ plugin. The key addition is the `AzureSkillsInitializer` class, which manages the lifecycle of Azure Skills using `npx` commands, and the corresponding registration in the plugin's XML configuration. It also adds a registry key to allow users to disable this functionality.

**New Azure Skills auto-install/update feature:**

* Added `AzureSkillsInitializer` class, which runs on project startup to check for Azure Skills installation and update status, performing a fresh install with `npx skills add` if not present, or updating with `npx skills update` if more than 24 hours have passed since the last update. The process is logged and can be disabled via a registry key.
* Registered `AzureSkillsInitializer` as a `postStartupActivity` in `azure-intellij-plugin-azuremcp.xml`, so it executes automatically after project startup.

**Configuration and user control:**

* Added a new registry key `azure.skills.autoconfigure.disabled` to allow users to disable the auto-install/update of Azure Skills, with a description and default value set to `false`.

No

Has this been tested?
---------------------------
- [x] Tested

### DIFF
--- a/PluginsAndFeatures/azure-toolkit-for-intellij/azure-intellij-plugin-azuremcp/src/main/java/com/microsoft/azure/toolkit/intellij/azuremcp/AzureSkillsInitializer.java
+++ b/PluginsAndFeatures/azure-toolkit-for-intellij/azure-intellij-plugin-azuremcp/src/main/java/com/microsoft/azure/toolkit/intellij/azuremcp/AzureSkillsInitializer.java
@@ -1,0 +1,244 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ */
+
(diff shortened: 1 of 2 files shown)


wangmingliang-ms:
Would automatically installing the skill on users' machines be inappropriate?
This approach sounds likely to provoke user complaints. I recommend offering the skill to users at an appropriate moment rather than installing it directly on their machines.

srnagar:
The skills will not be automatically enabled. The user still has to opt-in to use skills. So, in this change, we only set it up as part of the plugin startup and let users know that they have to opt-in to use skills. Also, before using each skill in the chat, there will be an elicitation request sent to the user to allow or deny using a skill.

wangmingliang-ms (approved):

PR #12293, opened 2026-08-16
Fix MicroBuild JAR signing regression

wangmingliang-ms:
## Summary
- pin `MicroBuild.Plugins.Signing` to the last known-good `1.1.1330` release
- pass `System.AccessToken` to plugin installation and JAR signing
- stop immediately when `DDSignFiles` returns a non-zero exit code

## Root cause
Scheduled real-sign builds started failing after the signing package automatically moved from `1.1.1330` to `1.1.1374`. The newer package looks for authentication certificate `ea4022a2-6f71-4214-88b6-18d2bbd6b4cc`, while PME installs `AUTH-CORP` with subject `dec434ad-6bd4-4a20-b400-4bf6db7fc3fb`.

The signing loop also ignored the failed process exit code, causing the task to retry every JAR until its four-hour timeout.

Failed build: https://mseng.visualstudio.com/VSJava/_build/results?buildId=31892392

## Validation
- `git diff --check`
- targeted checks for the pinned package, access-token propagation, fail-fast handling, and changed-file scope

### DIFF
--- a/.azure-pipelines/sign-for-stable-release.yml
+++ b/.azure-pipelines/sign-for-stable-release.yml
@@ -120,9 +120,12 @@ extends:
                     ConnectedPMEServiceName: "40012d40-9ded-4f75-8476-d60758200346"
                   ${{ else }}:
                     signType: "test"
+                  # Keep the last known-good package until the MicroBuild certificate regression is fixed.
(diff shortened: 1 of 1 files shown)


Flanker32 (approved):
LGTM

